PR #5, opened 2022-01-21
feat: use react

cbt1:
Removes the react-native implementation and replaces with a react implementation. No more is support for native platforms.


src/pivot-table/components/CellFactory.tsx
```diff
@@ -1,87 +1,52 @@
 import React from 'react';
-import { View, Text, StyleSheet } from 'react-native';
-import { Model } from '../../types/types';
-import { Cell, TYPE } from '../handle-data';
+import { ItemData } from '../../types/types';
+import { TYPE } from '../handle-data';
```
niekvanstaveren, line 3: TYPE looks like an enum. Not something that needs to be changed in this PR though. Also there might be a better name for it 😄

src/pivot-table/components/DimensionCell.tsx
```diff
@@ -1,26 +1,30 @@
...
+import { borderStyle, textStyle } from './shared-styles';
 
 export interface DimensionCellProps {
   cell: Cell;
   model: Model;
   rowIndex: number;
   colIndex: number;
   isLeftColumn?: boolean;
-  style: ViewStyle | ViewStyle[];
+  style: any;
```
niekvanstaveren, line 13: We should type this as well. Does this contain all the styling for a Cell?

src/pivot-table/components/DimensionCell.tsx
```diff
@@ -39,19 +43,15 @@ const DimensionCell = ({ model, cell, isLeftColumn = false, rowIndex = 0, colInd
...
-  const DimCell = (
-    <View style={style}>
-      <Text
-        style={[sharedStyles.text, styles.text]}
-        numberOfLines={1}>
-          {cellContent}
-      </Text>
-    </View>
+  return (
+    <div style={{ ...style, ...containerStyle}}>
```
niekvanstaveren, line 47: Would it make any sense creating a real table now that we don't do mobile native anymore?

src/types/stardust.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@nebula.js/stardust';
```
niekvanstaveren, line 1: We should check with Thomas on the state of stardust types being exported.

niekvanstaveren (approved):
Looks good from a quick view. Some notes that are not directly related to this PR, so can also be handled later

src/types/stardust.d.ts
cbt1: Agree. I just made dirty quick fix to remove the type error.

src/pivot-table/components/DimensionCell.tsx
cbt1: Not while we are using the `react-window` component. There each cell get an absolute position. Which allows us to only render part of a larger dataset.

src/pivot-table/components/DimensionCell.tsx
cbt1: It should be typed yes. It only contains the absolute position of the cell. Which is coming from the `react-window` component.

src/pivot-table/components/CellFactory.tsx
cbt1: Yeah it should be an enum. Not sure why I did do it like an object. Might have be that I ran into to some issue. Will take another look at it.

PR #6, opened 2022-01-26
feat: use constraints to disable interactions

cbt1:


niekvanstaveren (approved):
Adding some more people to look at what you are doing 😄

PR #7, opened 2022-01-26
feat: expand and collapse icon

cbt1:
Made an arbitrary choice of the icon

Screenshot 2022-01-26 at 10 08 50

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,10 @@
   "devDependencies": {
     "@babel/preset-react": "7.14.5",
     "@babel/preset-typescript": "7.15.0",
+    "@emotion/react": "11.7.1",
+    "@emotion/styled": "11.6.0",
+    "@mui/icons-material": "5.3.1",
+    "@mui/material": "5.3.1",
     "@nebula.js/cli": "2.3.1",
     "@nebula.js/cli-build": "2.3.1",
     "@nebula.js/cli-sense": "2.3.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,6 +30,10 @@
   "devDependencies": {
     "@babel/preset-react": "7.14.5",
     "@babel/preset-typescript": "7.15.0",
+    "@emotion/react": "11.7.1",
+    "@emotion/styled": "11.6.0",
```
niekvanstaveren, line 34: Are these necessary for this to work?

niekvanstaveren (approved):

package.json
cbt1: They are peerDeps to `@mui/material`

src/pivot-table/components/DimensionCell.tsx
```diff
@@ -65,23 +72,22 @@ const DimensionCell = ({
...
   }
 
   return (
     <div style={{ ...style, ...containerStyle}}>
       <div style={{ ...cellStyle, ...borderStyle }} onClick={onClickHandler} aria-hidden="true">
-        <div style={dimTextStyle}>
-            {cellContent}
-        </div>
+        {cellIcon}
+        <div style={dimTextStyle}>{qText}</div>
```
haxxmaxx, line 90: You should check out mui `styled` (not the same as styled components). Similar to `makeStyles` but supported by mui5. Then you don't have to do inline styling, which i think is slower. Some good examples in here: https://github.com/qlik-oss/sn-table/pull/426/files

haxxmaxx (approved):

src/pivot-table/components/DimensionCell.tsx
haxxmaxx: Not relevant to this PR though

cbt1: Agree, I will, at some point, have to take a look at styling. Including theming. But for now this will have to do.

PR #16, opened 2022-02-04
feat: sticky row/column headers

cbt1:
Adds support for sticky row/columns headers. Had to add a fair bit of complexity to make this work but it do seem to work out nicely. Would be nice to get to test this one day on a low performance device as well..

Screenshot 2022-02-04 at 10 35 49


src/pivot-table/components/DataGrid.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { memo, useCallback, useLayoutEffect } from "react";
+import { VariableSizeGrid, areEqual, GridOnItemsRenderedProps } from 'react-window';
+import { DataModel } from "../../types/types";
+import DataCell from './DataCell';
```
haxxmaxx, line 4: mixing " and ' here

cbt1: Yeah I need to fix the linting. I've missed semi-colon on a few places as well

src/pivot-table/data/extract-top.ts
```diff
@@ -0,0 +1,40 @@
...
+
+      if (node.qSubNodes.length) {
+        return extract(node.qSubNodes, mtrx, topRowIdx + 1);
+      }
+
+      return mtrx;
+    }, matrix);
+  };
+
+  return extract(qTop);
```
haxxmaxx, line 37: I would pass 0 here for clarity, makes it easier to understand the recursive part

src/pivot-table/data/extract-headers.ts
```diff
@@ -0,0 +1,25 @@
+import { NxDimensionInfo } from "../../types/QIX";
+import { Cell } from "../../types/types";
+import createCell from "./cell";
+
+const extractHeaders = (qDim: NxDimensionInfo[], rowCount: number, colCount: number): Cell[][] => {
+  const ary: Cell[][] = [];
+
+  for (let colIdx = 0; colIdx < colCount; colIdx += 1) {
+    for (let rowIdx = 0; rowIdx < rowCount; rowIdx += 1) {
```
haxxmaxx, line 9: `++` is nice here

src/pivot-table/components/__tests__/DimensionCell.test.tsx
```diff
@@ -32,7 +32,23 @@ describe('DimensionCell', () => {
     };
 
     dataModel = {
-      pivotData: { matrix: [], topMatrix: [], leftMatrix: [], nbrTopRows: 0, nbrLeftColumns: 0 },
+      pivotData: {
```
haxxmaxx, line 35: Would be nice to have this mocked if it is used in multiple places, not relevant for this pr though

src/pivot-table/components/DataGrid.tsx
haxxmaxx: OK I wont mind them then 👍

haxxmaxx:
Haven't looked super closely, but the 4 different grid components look very similar. Is it possible to make it into one general component?

src/pivot-table/components/__tests__/DataCell.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  test('should render null value',  () => {
+    cell.qText = '-';
+    cell.qNum = 'NaN';
+
+    render(<DataCell data={data} style={style} columnIndex={0} rowIndex={0} />);
+
+    expect(screen.getByText(cell.qText)).toBeInTheDocument()
+    expect(screen.getByTestId(testId).childNodes[0])
+      .toHaveStyle({ justifyContent: 'center', backgroundColor: '#f2f2f2' } as Record<string, unknown>);
```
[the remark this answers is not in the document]
cbt1: Yeah that's the only diff

src/pivot-table/components/__tests__/DimensionCell.test.tsx
cbt1: Agree and I've thought about it, but for now it's now that widely used so it will have to wait.

src/pivot-table/data/extract-headers.ts
cbt1: Again it's a lint rule that prevents me from doing that :(

src/pivot-table/data/extract-top.ts
cbt1: You mean something like this `extract(qTop, [], 0)` ?

cbt1:
> Haven't looked super closely, but the 4 different grid components look very similar. Is it possible to make it into one general component?

They are somewhat similar but also not and it will depend on how they evolves down the road. Currently working a data re-factoring which will set them apart a little bit more. But I will definitely keep that in mind.

PR #230, opened 2023-04-13
build: enable noUncheckedIndexedAccess

cbt1:
Enables the typescript compiler option `noUncheckedIndexedAccess` and fixes all ts errors caused by this


cbt1:
Lets not enable this for now

PR #231, opened 2023-04-14
fix: replace anti-pattern usuage of `useMemo` with `useOnPropsChange`

a-m-dev:
replacing direct useMemo usage in component body with a new hook hijacked from sn-table called `useOnPropsChange`


cbt1 (approved):

cbt1:
I would say that its a commit of type `refactor` rather than `fix` tho

a-m-dev:
✅ fixed, 
btw, I can't merge it because probably i don't have the right permissions to this repo 🙂

cbt1:
> ✅ fixed, btw, I can't merge it because probably i don't have the right permissions to this repo 🙂

How about now?

PR #234, opened 2023-04-18
feat: Calculate row height of column headers based on value from styling panel

a-m-dev:
This PR contains 2 sections:
- [x] add row height drop down into styling panel
- [x] do the math and update ui based on that 

Screenshot 2023-04-20 at 11 32 44


src/pivot-table/components/cells/DataCell.tsx
```diff
@@ -57,8 +57,8 @@ const MeasureCell = ({ columnIndex, rowIndex, style, data }: MeasureCellProps):
 
   return (
     <div title={text} style={{ ...style, ...containerStyle }} data-testid={testId}>
-      <div style={cellStyle}>
-        <span style={{ ...textStyle, ...styleService.content }}>{text}</span>
+      <div style={{ ...cellStyle, display: "flex" }}>
```
cbt1, line 60: FYI I do think we need to look at styling at some point but for now this is good enough

src/pivot-table/components/grids/LeftGrid.tsx
```diff
@@ -36,15 +37,13 @@ const rightListStyle: React.CSSProperties = {
   boxSizing: "content-box",
 };
 
-const DEFAULT_ROW_HEIGHT = 28;
-
-const getItemSizeCallback = (list: Cell[]) => (rowIndex: number) => {
+const getItemSizeCallback = (list: Cell[], defaultRowHeight: number) => (rowIndex: number) => {
```
cbt1, line 40: Minor suggestion but you might as well call this `cellHeight` instead

src/pivot-table/contexts/StyleProvider.tsx
```diff
@@ -12,8 +14,18 @@ const StyleContext = createContext<StyleService>(NOOP_STYLE_SERVICE);
 
 export const useStyleContext = (): StyleService => useContext(StyleContext);
 
-const StyleProvider = ({ children, styleService }: StyleProviderProps): JSX.Element => (
-  <StyleContext.Provider value={styleService}>{children}</StyleContext.Provider>
-);
+const StyleProvider = ({ children, styleService, layoutService }: StyleProviderProps): JSX.Element => {
```
cbt1, line 17: I do think we should expose cell heigth via the style provider. Not really sure about the "style service" and mixin it with props resolved here.

But once we support more styling options from the styling panel we should revisit this and figure a good way forward. Ex. is the any value in keeping the style service?

src/types/QIX.ts
```diff
@@ -35,10 +35,20 @@ interface NullValueRepresentation {
   text?: string;
 }
 
+export interface RowHeight {
+  linesCount: number;
+}
+
+export interface Component {
+  key: string;
```
cbt1, line 43: I think you should be explicit about the `key` here.

```suggestion
  key: "general" | "theme";
```

PR #236, opened 2023-04-21
feat: line clamp for header cell

a-m-dev:
This PR fixes Line clamp for header cells, basically the idea is to tie row height we get from styling panel to the lines we show in header (dim) cells. 

Below is a quick demo of what will happen: 

https://user-images.githubusercontent.com/29652890/233600222-d71049bc-5795-4381-88be-10a2b46ddf83.mov

### DIFF
--- a/src/pivot-table/components/cells/DimensionCell.tsx
+++ b/src/pivot-table/components/cells/DimensionCell.tsx
@@ -44,11 +44,20 @@ const stickyCell: React.CSSProperties = {
   top: 4,
 };
 
-const dimTextStyle: React.CSSProperties = {
+const getDimTextStyle = (isLeftColumn: boolean, clampCount: number): React.CSSProperties => ({
   ...textStyle,
+  whiteSpace: "unset",
   fontWeight: "bold",
   marginLeft: 4,
-};
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+  display: "-webkit-box",
+  ...(!isLeftColumn && {
+    lineClamp: clampCount,
+    WebkitLineClamp: clampCount,
+    WebkitBoxOrient: "vertical",
+  }),
+});
 
 const nullStyle: React.CSSProperties = {
   backgroundColor: "#f2f2f2",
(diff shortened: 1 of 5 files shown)


cbt1:
How does it line break if you have a single word that is really really long without any white space?

a-m-dev:
it works well when we have `line-break: anywhere` with a very long text without space:
Screenshot 2023-04-21 at 12 44 13

which we have full support even in safari:
Screenshot 2023-04-21 at 12 46 17

will update PR soon

cbt1:
How about pseudo dimensions? When you have multiple measures acting as a dimension that is. If you have a long label. Will it line break? Looks like you have only modified the dimension cell component

src/pivot-table/components/cells/DimensionCell.tsx
```diff
@@ -44,11 +44,21 @@ const stickyCell: React.CSSProperties = {
...
   ...textStyle,
+  whiteSpace: "unset",
   fontWeight: "bold",
   marginLeft: 4,
-};
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+  display: "-webkit-box",
+  ...(!isLeftColumn && {
+    lineClamp: clampCount,
```
cbt1, line 56: Does the `line-clamp` actual exist? Cannot find it anywhere on MDN

src/pivot-table/components/cells/DimensionCell.tsx
```diff
@@ -44,11 +44,21 @@ const stickyCell: React.CSSProperties = {
...
+const getDimTextStyle = (isLeftColumn: boolean, clampCount: number): React.CSSProperties => ({
   ...textStyle,
+  whiteSpace: "unset",
   fontWeight: "bold",
   marginLeft: 4,
-};
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+  display: "-webkit-box",
+  ...(!isLeftColumn && {
```
cbt1, line 55: Does this mean that line break only happens for the "top grid"?

src/pivot-table/components/cells/DimensionCell.tsx
a-m-dev: My assumption was passing it there + vendor prefix one with webkit (since apparently firefox supports it as well) and when it becomes stable (without vendor prefix) we are going to already have it there. 

But now, I think probably we don't need it. will remove it and only keep webkit one!

cbt1 (approved):
Did some testing on this, looks like it's working nicely 👍

PR #240, opened 2023-04-24
feat: line clamp for grid cells

a-m-dev:
line clamp for grid cells

https://user-images.githubusercontent.com/29652890/233957126-dd3ea39a-3fda-4d5c-8bb6-81458e7626af.mov

### DIFF
--- a/src/pivot-table/components/cells/DataCell.tsx
+++ b/src/pivot-table/components/cells/DataCell.tsx
@@ -3,7 +3,8 @@ import { areEqual } from "react-window";
 import NxDimCellType from "../../../types/QIX";
 import { GridItemData } from "../../../types/types";
 import { useStyleContext } from "../../contexts/StyleProvider";
-import { borderStyle, textStyle } from "../shared-styles";
+import includesDigit from "../helpers/includes-digit";
+import { borderStyle, getLineClampStyle, textStyle } from "../shared-styles";
 
 export interface MeasureCellProps {
   columnIndex: number;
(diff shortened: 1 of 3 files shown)


src/pivot-table/components/cells/DataCell.tsx
```diff
@@ -52,13 +59,23 @@ const MeasureCell = ({ columnIndex, rowIndex, style, data }: MeasureCellProps):
...
   return (
     <div title={text} style={{ ...style, ...containerStyle }} data-testid={testId}>
-      <div style={{ ...cellStyle, display: "flex" }}>
-        <span style={{ ...textStyle, ...styleService.content, alignSelf: "flex-start" }}>{text}</span>
+      <div style={cellStyle}>
+        <span
+          style={{
+            ...textStyle,
+            ...styleService.content,
+            ...(!includesDigit(text) && getGridTextClampStyle(styleService.lineClamp)),
```
cbt1, line 73: We should use line clamp if the value is not numeric. So this should do it:
```suggestion
            ...(!isNumeric && getGridTextClampStyle(styleService.lineClamp)),
```

a-m-dev: it actually does not work as what we discussed, if we have number in text -> it does not limit lines back to only 1:

https://user-images.githubusercontent.com/29652890/233991850-72c87d43-2b73-440a-bd55-c1a32b2cb34a.mov

cbt1: I don't recall use saying that if it's contains it a number we should not line break it.

We want it to behave like in the sn-table. So if the cell value is a numeric value, no line break is allow. If the values is not a numeric, line break is allowed. Which should be what my suggested code does.

PR #241, opened 2023-04-24
chore: api spec update

a-m-dev:
for PT

### DIFF
--- a/src/qae/initial-properties.d.ts
+++ b/src/qae/initial-properties.d.ts
@@ -1,35 +1,35 @@
 import { Component } from "../types/QIX";
 
-declare module "initial-properties" {
-  declare const properties: {
-    version: number;
-    qHyperCubeDef: {
-      qAlwaysFullyExpanded: boolean;
-      qDimensions: EngineAPI.INxDimension[];
-      qMeasures: EngineAPI.INxMeasure[];
-      qMode: "P";
-      qSuppressMissing: boolean;
-      qSuppressZero: boolean;
-      qShowTotalsAbove: boolean;
-      qInitialDataFetch: [
-        {
-          qTop: 0;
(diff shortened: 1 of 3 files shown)


a-m-dev:
why api-governance step keeps failing? any idea about it?

cbt1:
> yes, you have updated the api-specification without changing the version. You should only re-generate the api specs when making a release

My bad, I told him to update it.

src/qae/initial-properties.d.ts
```diff
@@ -1,32 +1,35 @@
-declare module "initial-properties";
+import { Component } from "../types/QIX";
```
cbt1, line 1: `Component` does not exist. You need to add it

a-m-dev: I actually followed what we do in table side: https://github.com/qlik-oss/sn-table/blob/main/src/qae/object-properties.d.ts#L1

is that one correct or should I declare something like: 
```ts
declare const RowHeight {
  linesCount: number;
}

declare const Component {
  key: "general" | "theme";
  rowHeight?: RowHeight;
}
```

a-m-dev: Actually, any declarations in `*.d.ts` file is only for our local development, for our project and internal usage of typescript to not make it confuse with types. 
So the thing is `Component` interface is already declared inside of `types/QIX` and if we are using this as internal (more) type interference of typescript, it makes sense to give it what it needs (import the `Component` interface and use it in `properties.component`)

cbt1: Looks like I was on a branch that was not up-to-date with `main` so I did not have `Component` defined

cbt1 (approved):

PR #242, opened 2023-04-25
feat: wait for fonts

cbt1:
End up needing this in #237 

It's also a prime candidate for being a shared hook between sn-table and sn-pivot-table.


cbt1:
Rendering baseline updated as the column width has changed now when `useMeasureText` uses the correct font

haxxmaxx (approved):

a-m-dev (approved):
LGTM

PR #243, opened 2023-04-26
Crl/styling

haxxmaxx:


src/pivot-table/contexts/create-styling.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * creates the styling based on layout, theme and default values - in that order
+ */
+export const createStyling = (layoutService: LayoutService, theme: ExtendedTheme): Styling => {
+  1;
+  const chartStyling = layoutService.layout.components?.find((n) => n.key === "theme");
+
+  const lineClamp =
+    chartStyling?.rowHeight?.linesCount ??
+    // theme.getStyle(BASE_PATH, Path.ROOT, Attribute.LINE_CLAMP) ?? // TODO: add line clamp to theme
```
cbt1, line 71: Might need to create a new theme path for the sn-pivot-table if we add line clamp to the theme. Same as we did for sn-table

src/ext/property-panel/settings.ts
```diff
@@ -24,6 +24,100 @@ const getStylingPanelConfig = () => ({
...
+                      translation: "properties.fontSize",
+                      width: 9,
+                      min: 5,
+                      max: 300,
+                      defaultValue(item, data, args) {
+                        const currentTheme = args.theme.current();
+                        return parseInt(currentTheme.object?.pivotTable?.header?.fontSize ?? currentTheme.fontSize, 10);
+                      },
+                      change(data) {
+                        data.header.fontSize = !data.header.fontSize
```
cbt1, line 51: If I input a non-numerical value, the behaviour is kind of strange. The value is removed and replace but "undefined" as a placeholder

haxxmaxx: yeah this logic looks weird. It is the same as for table. Maybe the logic worked better with the old styling table?

src/pivot-table/contexts/StyleProvider.tsx
```diff
@@ -1,32 +1,23 @@
...
-export const useStyleContext = (): StyleService => useContext(StyleContext);
+export const useStyleContext = (): Styling => useContext(StyleContext);
 
-const StyleProvider = ({ children, styleService, layoutService }: StyleProviderProps): JSX.Element => {
-  const rowHeight = useMemo(
-    () => layoutService.layout.components?.find((n) => n.key === "theme")?.rowHeight,
-    [layoutService.layout.components]
-  );
+const StyleProvider = ({ children, theme, layoutService }: StyleProviderProps): JSX.Element => {
+  const memoisedStyling = useMemo(() => createStyling(layoutService, theme), [layoutService, theme]);
```
cbt1, line 18: Won't be re-created when theme changes

```suggestion
  const memoisedStyling = useMemo(() => createStyling(layoutService, theme), [layoutService, theme.name()]);
```

PR #244, opened 2023-04-26
feat: cell height based on font size

a-m-dev:
dynamically calcualte cell height based on font size (from theme) and row height we get from styling panel

https://user-images.githubusercontent.com/29652890/234535027-0b541fb9-7d3a-47ef-8e09-7cd761f357fe.mov


src/pivot-table/hooks/use-cell-height.ts
```diff
@@ -0,0 +1,35 @@
...
+    contentCellHeight: number;
+  };
+}
+
+const LINE_HEIGHT_COEFFICIENT = 4 / 3;
+const fontSizeToRowHeight = (fontSize: string, lineClamp: number) =>
+  +(parseInt(fontSize, 10) * LINE_HEIGHT_COEFFICIENT * lineClamp).toFixed(2);
+
+const useCellHeight: UseCellHeight = ({ styleService, layoutService }) => {
+  const lineClamp = useMemo<number>(() => {
```
haxxmaxx, line 23: I will probably rewrite this a bit in my PR :)

src/pivot-table/hooks/use-cell-height.ts
```diff
@@ -0,0 +1,35 @@
...
+  (args: UseCellHeightProps): {
+    lineClamp: number;
+    headerCellHeight: number;
+    contentCellHeight: number;
+  };
+}
+
+const LINE_HEIGHT_COEFFICIENT = 4 / 3;
+const fontSizeToRowHeight = (fontSize: string, lineClamp: number) =>
+  +(parseInt(fontSize, 10) * LINE_HEIGHT_COEFFICIENT * lineClamp).toFixed(2);
```
haxxmaxx, line 20: Is this rounding from sn-table? If so, there might be a bug caused by it. Let's keep our eyes open if borders disappear for certain font sizes

a-m-dev: not actually, since we have `LINE_HEIGHT_COEFFICIENT` we sometimes end up in numbers like 48.3333333334. and those numbers were causing some issues in test => so i decided to only round it with 2 digits (if it was float number).

but sure, we can consider that possible bugs as well

haxxmaxx: yeah there is something similar in the table, and I suspect we have a bug because of it. But if everything looks fine for now, no need to change anything id say

src/pivot-table/hooks/use-cell-height.ts
```diff
@@ -0,0 +1,35 @@
...
+interface UseCellHeight {
+  (args: UseCellHeightProps): {
+    lineClamp: number;
+    headerCellHeight: number;
+    contentCellHeight: number;
+  };
+}
+
+const LINE_HEIGHT_COEFFICIENT = 4 / 3;
+const fontSizeToRowHeight = (fontSize: string, lineClamp: number) =>
```
cbt1, line 19: I think you need to include cell padding here. I did some testing and with characters like "g", it can get clipped at the bottom.

cbt1:
You should set `lineHeight: calc(4/3)` on all cells so that they render with the line height we are using for calculating the cell height

src/pivot-table/hooks/use-cell-height.ts
cbt1: I would have actually tought that `.toFixed(2)` helps to mitigate the floating precision issue

cbt1 (changes requested):
See comments regarding padding and lineHeight on cell level

src/pivot-table/contexts/__tests__/StyleProvider.test.tsx
```diff
@@ -26,18 +27,26 @@ describe("StyleProvider", () => {
       </StyleProvider>
     );
 
+  // formula:
+  // given fontsize * coefficient (most adaptive one) * clamp count + padding of 8px
+  const fontSizeCalcualteHelper = (size: string, clamp: number): string =>
+    `${+(parseInt(size, 10) * (4 / 3) * clamp + 8).toFixed(2)}`;
```
cbt1, line 33: You can use the constant for padding here

cbt1 (approved):

cbt1 (changes requested):
Looks like some background styling sneaked into the rendering baselines

a-m-dev:
Just fixed it
Since we using an extension build for rendering tests, my build was a bit old and based on my previous debugging steps.

cbt1 (approved):

PR #245, opened 2023-04-26
fix: expand or collapse data fetch

cbt1:
When a node was expanded or collapsed, all data before it was fetched again so that the scroll position would not be lost. But this would fetch +100 000 rows if a user had scroll to row 100 000 and expanded that node. So not ideal.

This PR removes that code since it is no longer needed after #237 to keep the scroll position. If the user scroll, for example row 100 000 and expand that node. The scroll position is keep there but all those 100k rows above are empty until the user scroll up. At which point the nodes currently rendered on screen are fetched.


a-m-dev (approved):
🚀

PR #247, opened 2023-04-27
feat: limit column count

cbt1:
Added an artificial column limit of 75K. I did some testing with the minimal column width and that is where it ended up. It need to change into a dynamic value later on. One that is based on the estimated colum width and limited by some fixed max element width value (ex. 10M px).

Also ended up doing tweaking on how "sizes" are exposed:
- layout service now contains the size of the layout (limited value)
- Left- and Top- dimension data now contains the size of its data (row and column count). Before it return both x/y but only one of them was used.
- `isSnapshot` is now only resolved once
- `measureData` has been simplified and now only contains raw data


src/services/layout-service.ts
```diff
@@ -1,12 +1,15 @@
...
 import { NxSelectionCellType, PivotLayout } from "../types/QIX";
 import { LayoutService } from "../types/types";
 
 const createLayoutService = (layout: PivotLayout): LayoutService => {
-  const { qHyperCube, nullValueRepresentation } = layout;
+  const { qHyperCube, nullValueRepresentation, snapshotData } = layout;
   const { qNoOfLeftDims, qEffectiveInterColumnSortOrder, qMeasureInfo, qDimensionInfo } = qHyperCube;
   const leftDimensions = qDimensionInfo.slice(0, qNoOfLeftDims);
   const topDimensions = qDimensionInfo.slice(qNoOfLeftDims);
+  const isSnapshot = !!layout.snapshotData;
```
a-m-dev, line 11:
```suggestion
  const isSnapshot = !!snapshotData;
```
maybe?

a-m-dev (approved):
LGTM

PR #248, opened 2023-04-27
feat: styling from layout

haxxmaxx:
Moving all styling logic to the style service. Now the all the provider file does is add the styleService to the context. Also adding styling changes from layout. The prioriety is layout first, then theme, then default. For some props there isn't all three steps.

- background: only default. We maybe should set it using the setting under the general tab. Also no way of doing it using the theme API AFIAK
- lineClamp. Not is theme yet, we need to descide on a name
- fontFamily. Not in styling panel yet

Colors are not dealt with properly yet, will be the next PR


src/types/QIX.ts
```diff
@@ -39,9 +39,49 @@ export interface RowHeight {
...
+    fontColor?: PaletteColor;
+  };
+  content?: {
+    fontFamily?: string;
+    fontSize?: number;
+    fontColor?: PaletteColor;
+  };
+}
+
+export interface CurrentTheme {
```
cbt1, line 61: Not sure if this fits into `QIX` types. But lets keep it here for now and later figure out with it means to be a qix type

cbt1:
I'm thinking that we can just remove `background`, it does not exist on a chart level in the theme. Can't really remember why it was added in the first place but looking at it now, I don't think we need it

cbt1 (approved):
Code LGTM, we can decide what to do with background later

a-m-dev (commented):
LGTM

a-m-dev (approved):

PR #249, opened 2023-04-27
feat: limited data disclaimer

cbt1:
Adds a disclaimer at the bottom when there is more column data available then what can be rendered.

TODO:s
- Disclaimer string should be translated
- Font family should be resolved from root of theme to match the other charts styling for disclaimer text

Skärmavbild 2023-04-27 kl  14 23 13


src/pivot-table/components/Wrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { PivotTableProps, StickyPivotTable } from "./PivotTable";
+
+export const Wrapper = (props: PivotTableProps): JSX.Element => {
+  const {
+    layoutService: { hasLimitedData },
+  } = props;
+  const styleService = useStyleContext();
+
+  return (
+    <div>
```
a-m-dev, line 13: I think fragment should be a good choice here

src/pivot-table/hooks/use-table-rect.ts
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from "react";
+import { LayoutService, Rect } from "../../types/types";
+import { DISCLAIMER_HEIGHT } from "../constants";
+
+export const useTableRect = (rect: Rect, layoutService: LayoutService) =>
+  useMemo(
+    () => (layoutService.hasLimitedData ? { ...rect, height: rect.height - DISCLAIMER_HEIGHT } : rect),
```
a-m-dev, line 7: I think probably the usage of `layoutService.hasLimitedData` + `DISCLAIMER_HEIGHT` subtraction from height, will have side effect on height calculations we do in PivotTable component 

https://github.com/qlik-oss/sn-pivot-table/blob/f2afd5f8f9c1554ebc5064e575188fa665c37959/src/pivot-table/components/PivotTable.tsx#L102-L107

a-m-dev (commented):
code looks LGTM, just a few concerns

src/pivot-table/components/Wrapper.tsx
cbt1: Fixed

a-m-dev (approved):

PR #251, opened 2023-04-28
feat: locale

cbt1:
Adds locale and let the disclaimer consume it


src/hooks/use-translations.ts
```diff
@@ -0,0 +1,12 @@
+import { useMemo, useTranslator } from "@nebula.js/stardust";
+import registerLocale from "../locale/src";
+import { ExtendedTranslator } from "../types/types";
+
+export const useTranslations = () => {
+  const translator = useTranslator() as ExtendedTranslator;
+  const language = translator.language();
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useMemo(() => registerLocale(translator), [translator, language]);
```
haxxmaxx, line 9: why memo and not effect?

haxxmaxx (approved):
Hopefully we don't need the translator as much as in the table. Then we need to pass it around alot and would probably be put on some kind of context thing

src/hooks/use-translations.ts
cbt1: Actually no real reason, just ended up being memo. But I guess you could end up in a scenario when using "useEffect" where the chart is first rendered without having locale registered. As "useEffect" runs after rendering. Not likely to happen tho.

haxxmaxx: ok I just never thought of using it this way, since we are not saving a value

haxxmaxx: ah yes we do it inside a useMemo in sn-table too. But we actually return a value there. This should be fine

haxxmaxx (approved):

mattgengqlik:
@haxxmaxx @cbt1  Is `src/locale/locales/en.json` a new file that we need to start translating?

haxxmaxx:
> @haxxmaxx @cbt1 Is `src/locale/locales/en.json` a new file that we need to start translating?

yes

src/locale/locales/en.json
```diff
@@ -0,0 +1,6 @@
+{
+  "SNPivotTable.LimitedData": {
+    "value": "* Currently showing a limited number of columns",
```
mattgengqlik, line 3: @JuliaBeingessner Could you help review the English strings when you have time?

cbt1:
> @haxxmaxx Could you help get @JuliaBeingessner `write` access to the repo? She is the tech writer who will review the English strings. Thanks

Done @JuliaBeingessner @mattgengqlik

src/hooks/use-translations.ts
a-m-dev: I think we can even skip this useMemo here, Because: 

1. we [have a guard inside of `registerLocale()`](https://github.com/qlik-oss/sn-pivot-table/blob/6ba1abbd1824b66b54de04178809a0c9e52bfee7/src/locale/src/index.ts#L12)
2. in every function call of `useTranslations()` -> we will hit that guard and return immediately from it

The fact that `useMemo` will only skip re-triggering when that hook/component is being re-called/rendered => but the ratio of being re-triggered (in order to memoise) VS it all starts from beginning is less and even in normal scenarios, we are going to hit that guard and do an immediate return and having a Memo on registering it, is kinda overkill IMO

a-m-dev (approved):

mattgengqlik (approved):
LGTM

PR #253, opened 2023-05-02
fix: borders

cbt1:
General overhaul of borders. 

- Fixes an issue where overlapping borders would make them a bit fuzzy
- Adjust the borders to new UX designs
  - Border color
  -  Removal of border around the chart
- Simplifies the "Header grid", as it did not need be rendered as a react-window component
  - Data logic needs to be updated as well, but will do that in a separate PR


a-m-dev (approved):
Code looks LGTM!

PR #254, opened 2023-05-03
ci: locale verify job

cbt1:


package.json
```diff
@@ -36,7 +36,7 @@
     "prepare": "husky install",
     "sense": "nebula sense",
     "spec": "scriptappy-from-jsdoc -c ./scriptappy.config.js",
-    "start": "nebula serve  --type sn-pivot-table",
+    "start": "yarn run locale:generate && nebula serve  --type sn-pivot-table",
```
haxxmaxx, line 39: do we need it here?

src/hooks/use-translations.ts
```diff
@@ -1,12 +1,12 @@
...
 import registerLocale from "../locale/src";
 import { ExtendedTranslator } from "../types/types";
 
 export const useTranslations = () => {
   const translator = useTranslator() as ExtendedTranslator;
   const language = translator.language();
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  useMemo(() => registerLocale(translator), [translator, language]);
+
+  useEffect(() => registerLocale(translator), [translator, language]);
```
haxxmaxx, line 9: how come you change to useEffect?

package.json
cbt1: Maybe, but I guess it depends on how you use the scripts.

If, for example, the first thing you do when having cloned the repo, is `yarn` and than `yarn start` we want it to contain the generated locale.

I myself don't use the build scripts as `yarn start` do the watch and building as well. But if I changed a locale file, it probably run the generate script myself.

src/hooks/use-translations.ts
cbt1: I did kind of agreed with https://github.com/qlik-oss/sn-pivot-table/pull/251#discussion_r1182177396 and it was nice to get ride of the disabled lint rule. But having giving it some more thought.

It boils down to which `useEffect` finish first, the one that register locale OR the `useEffect` that render the chart, as they are both async. While I do think it's unlikely that registering locale will take more time to finish than rendering the actual chart. We simply do know what `translator.add()` calls does and how much work that requires.

So I will actually reverat back to `useMemo` again.

@a-m-dev

package.json
haxxmaxx: oh I didn't know start does building as well, then it makes sense

haxxmaxx (approved):

PR #255, opened 2023-05-05
fix: dimension title cell key

cbt1:
Extract the dimenssion title cell key from qDimensionInfo instead of using column index.

### DIFF
--- a/src/pivot-table/data/extract-headers.ts
+++ b/src/pivot-table/data/extract-headers.ts
@@ -1,19 +1,23 @@
 import { PSEUDO_DIMENSION_INDEX } from "../../constants";
+import type { ExtendedDimensionInfo } from "../../types/QIX";
+import type { HeaderTitle } from "../../types/types";
 
 const extractHeaders = (
   qDim: EngineAPI.INxDimensionInfo[],
   rowCount: number,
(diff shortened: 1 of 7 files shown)


src/pivot-table/data/extract-headers.ts
```diff
@@ -1,19 +1,23 @@
...
-  const matrix: (null | string)[][] = Array(dimensionInfoIndex.length)
+): (null | HeaderTitle)[][] => {
+  const matrix: (null | HeaderTitle)[][] = Array(dimensionInfoIndex.length)
     .fill(null)
     .map(() => Array.from({ length: rowCount }, () => null));
 
   dimensionInfoIndex.forEach((dimIndex, colIdx) => {
     if (dimIndex === PSEUDO_DIMENSION_INDEX) {
-      matrix[colIdx][rowCount - 1] = "";
+      matrix[colIdx][rowCount - 1] = { id: "PSEUDO-DIM", title: "" };
```
a-m-dev, line 16: cell.id is been used in `DimensionTitleCell` and wouldn't this cause duplicate key error in react? in case if we have multiple pseudo dim?

a-m-dev (approved):

PR #257, opened 2023-05-08
feat: pagination

a-m-dev:
pagination feature, including:

- moving page info to supernova level
- introducing `usePVData`
- introducing `Pagination` component
- current `MAX_COL/MAX_ROW` is set to 15K (fair performance when we have both x and y in 15K cells length)


src/pivot-table/data/__tests__/use-pv-data.test.ts
```diff
@@ -116,21 +147,19 @@ describe("useData", () => {
...
-    expect(result.current.leftDimensionData).toBe(leftDimensionData);
+    expect(result.headersData).toBe(headersData);
+    expect(result.measureData).toBe(measureData);
+    expect(result.topDimensionData).toBe(topDimensionData);
+    expect(result.leftDimensionData).toBe(leftDimensionData);
   });
 
-  test("calling nextPageHandler should trigger data updates", () => {
-    const { result } = renderHook(() => useData(qPivotDataPages, layoutService));
+  test.skip("calling nextPageHandler should trigger data updates", () => {
```
a-m-dev, line 158: in this test, what is happening is that `result.nextPageHandler(...)`should trigger a state change. Based on that state change -> an `useEffect` might be called, and we are asserting the function calls within that `useEffect` (which are mostly mocked), but since we are mocking the `useEffect` itself (just calling the callback that is being passed into it, not recall that callback in every dependency check) this becomes impossible to handle || at least i don't have any idea to how to handle it. 

Any suggestions are welcome on this one 🙏

src/hooks/use-pivot-table.ts
```diff
@@ -0,0 +1,121 @@
...
+  language,
+}: UsePivotTableProps) => {
+  const {
+    size,
+    layout: {
+      qHyperCube: { qSize },
+    },
+  } = layoutService;
+
+  const rowsPerPage = qSize.qcy <= MAX_ROW_COUNT ? qSize.qcy : MAX_ROW_COUNT;
```
cbt1, line 55: You should be able to do:
```suggestion
  const rowsPerPage = Math.min(qSize.qcy, MAX_ROW_COUNT):
```

cbt1, line 55: You should be able to do:
```suggestion
  const rowsPerPage = Math.min(qSize.qcy, MAX_ROW_COUNT):
```

src/pivot-table/data/use-pv-data.ts
```diff
@@ -0,0 +1,100 @@
+import { useEffect, useMemo, useState } from "@nebula.js/stardust";
```
cbt1, line 1: This is a Nebula custom hook and should be placed under `src/hooks/` with the other Nebula hooks

src/pivot-table/data/use-pv-data.ts
```diff
@@ -0,0 +1,100 @@
...
+import { addPageToMeasureData, createMeasureData } from "./measure-data";
+import { addPageToTopDimensionData, createTopDimensionData } from "./top-dimension-data";
+
+interface UsePVDataProps {
+  qPivotDataPages: EngineAPI.INxPivotPage[];
+  layoutService: LayoutService;
+  pageInfo: PageInfo;
+}
+
+export const usePVData = ({ qPivotDataPages, layoutService, pageInfo }: UsePVDataProps): Data => {
```
cbt1, line 22: Is this hook really needed? Can you not use `useLoadDataPages` to get the "initial" data for a given page? That hook just need to know the page that is currently active to load data

a-m-dev:
closing this one in favour of https://github.com/qlik-oss/sn-pivot-table/pull/275

a-m-dev:
closing this one in favour of https://github.com/qlik-oss/sn-pivot-table/pull/275

PR #258, opened 2023-05-08
fix: resolve palette colors

haxxmaxx:
uses the theme to resolve color, making sure we pick the color from palette when there is a valid index

### DIFF
--- a/src/types/QIX.ts
+++ b/src/types/QIX.ts
@@ -40,8 +40,8 @@ export interface RowHeight {
 }
 
 export interface PaletteColor {
-  index?: number | undefined;
-  color?: string | undefined;
+  index?: number;
+  color?: string;
 }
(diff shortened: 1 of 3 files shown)


src/types/QIX.ts
```diff
@@ -40,8 +40,8 @@ export interface RowHeight {
 }
 
 export interface PaletteColor {
-  index: number;
-  color: string | null;
+  index?: number | undefined;
```
cbt1, line 43: No need for `| undefined` when the property is optional

cbt1 (approved):

src/types/QIX.ts
haxxmaxx: I'm using the same typing from the stardust api, but It works fine without ` | undefined` so I will remove it

PR #266, opened 2023-05-09
refactor: fix dupl warnings

haxxmaxx:
I took a look at the duplicate warnings on code climate. There is generally a lot of similar code, a lot due to some of the grids having do to the same things, but with different qTop/qLeft x/y row/col stuff. No real easy way of generalizing without making it hard to read.

We could potentially raise the simiilar-code threshold. But I mean it depends on what we want code climate to do. Catch all errors including things we don't want to change, or have a looser check that might miss some things that actually needs refactoring?

### DIFF
--- a/src/pivot-table/hooks/use-selections-model.ts
+++ b/src/pivot-table/hooks/use-selections-model.ts
@@ -40,17 +40,17 @@ export default function useSelectionsModel(selections: ExtendedSelections): Sele
     (qType: EngineAPI.NxSelectionCellType, qRow: number, qCol: number) => {
       switch (qType) {
         case NxSelectionCellType.NX_CELL_LEFT:
-          return !!selected.find((cell) => {
-            if (cell.qType === NxSelectionCellType.NX_CELL_TOP) return true;
-            if (cell.qType === NxSelectionCellType.NX_CELL_LEFT && cell.qCol !== qCol) return true;
-            return false;
-          });
+          return !!selected.find(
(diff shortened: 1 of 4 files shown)


src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -40,17 +40,17 @@ export default function useSelectionsModel(selections: ExtendedSelections): Sele
     (qType: EngineAPI.NxSelectionCellType, qRow: number, qCol: number) => {
       switch (qType) {
         case NxSelectionCellType.NX_CELL_LEFT:
-          return !!selected.find((cell) => {
-            if (cell.qType === NxSelectionCellType.NX_CELL_TOP) return true;
-            if (cell.qType === NxSelectionCellType.NX_CELL_LEFT && cell.qCol !== qCol) return true;
-            return false;
-          });
+          return !!selected.find(
```
haxxmaxx, line 43: this one didn't fix the problem, I can revert it if we prefer the old way

cbt1:
> We could potentially raise the simiilar-code threshold. But I mean it depends on what we want code climate to do. Catch all errors including things we don't want to change, or have a looser check that might miss some things that actually needs refactoring?

IMO it's fine that it warns about issues. If it becomes unreasonable we may want to tweak the settings for it, otherwise we can just mark it as "Won't fix" if we are fine with the code?

a-m-dev (approved):

PR #271, opened 2023-05-10
refactor: move all data fetching into supernova level

a-m-dev:
by this PR data fetching only happen in super nova level, except fetching for more data which happens inside of `useData` hook in react level


src/hooks/use-load-data-pages.ts
```diff
@@ -23,9 +23,33 @@ const shouldFetchAdditionalData = (
...
+const isMissingLayoutData = (layoutService: LayoutService): boolean => {
+  const { qHyperCube } = layoutService.layout;
+  // if qPivotDataPages is array
+  if (Array.isArray(qHyperCube.qPivotDataPages)) {
+    if (
+      // if qPivotDataPages not contains data
+      qHyperCube.qPivotDataPages.length === 0 ||
+      // if qPivotDataPages has data but it is empty
+      qHyperCube.qPivotDataPages[0].qData.length === 0 ||
+      // ???
```
a-m-dev, line 35: @cbt1 is this translates to fetch more on x axis? 
and below is fetch more as much as we can (within the boundary of a page height) in y axis?

cbt1: I would say no. We make an assumption there that we need to data on X and Y axis if any of the axeses are missing data

src/hooks/use-load-data-pages.ts
```diff
@@ -23,9 +23,33 @@ const shouldFetchAdditionalData = (
   );
 };
 
+const isMissingLayoutData = (layoutService: LayoutService): boolean => {
```
cbt1, line 26: I'm thinking this could be simplified into:

```js
const isMissingLayoutData = (layoutService: LayoutService): boolean => {
  const { qHyperCube } = layoutService.layout;
  const { qWidth, qHeight } = qHyperCube.qPivotDataPages[0]?.qArea ?? { qWidth: 0, qHeight: 0 };

  return qWidth < DEFAULT_PAGE_SIZE || qHeight < DEFAULT_PAGE_SIZE;
};
```

src/hooks/use-load-data-pages.ts
```diff
@@ -35,29 +59,34 @@ const useLoadDataPages = (model: Model, layoutService: LayoutService, viewServic
   usePromise(async () => {
```
cbt1, line 59: You need to first check if it is a snapshot and if so use the data page from that.

Example:
```js
usePromise(async () => {
    if (layoutService.isSnapshot) {
      setDataPages(layoutService.layout.snapshotData?.content?.qPivotDataPages || []);
    } else if (
      model &&
      (shouldFetchAdditionalData(qLastExpandedPos, viewService) || isMissingLayoutData(layoutService))
    ) {
      try {
        const pivotPages = await (model as EngineAPI.IGenericObject).getHyperCubePivotData(Q_PATH, [
          {
            qLeft: qLastExpandedPos ? viewService.gridColumnStartIndex : 0,
            qTop: qLastExpandedPos ? viewService.gridRowStartIndex : 0,
            qWidth: !viewService.gridWidth ? DEFAULT_PAGE_SIZE : viewService.gridWidth,
            qHeight: !viewService.gridHeight ? DEFAULT_PAGE_SIZE : viewService.gridHeight,
          },
        ]);

        setDataPages([...qHyperCube.qPivotDataPages, ...pivotPages]);
      } catch (error) {
        // TODO handle error
        console.error(error);
      }
    } else if (qHyperCube.qPivotDataPages.length) {
      // when we get new page
      setDataPages(qHyperCube.qPivotDataPages);
    }

    ref.isLoading = false;
  }, [layoutService.layout, model, viewService, snapshotData]);
```

src/locale/all.json
```diff
@@ -0,0 +1,22 @@
+{
```
cbt1, line 1: This file should not be commited. Delete it and than merge latest from `main`

src/hooks/use-load-data-pages.ts
cbt1: Note that I've updated the example so it should now work with rendering tests

src/locale/all.json
```diff
@@ -0,0 +1,22 @@
...
+  "SNPivotTable_LimitedData": {
```
cbt1, line 2: Can you delete this file?

a-m-dev: I actually did, but weird, let me check it again

cbt1 (approved):

PR #273, opened 2023-05-11
refactor: `useRender()`

a-m-dev:


src/index.ts
```diff
@@ -1,27 +1,9 @@
 /* eslint-disable react-hooks/rules-of-hooks */
```
cbt1, line 1: Can remove this line

cbt1 (approved):

src/index.ts
a-m-dev: not actually, eslint is still confused about react env and supernova env

Screenshot 2023-05-11 at 15 14 09

PR #274, opened 2023-05-11
refactor: item size handler

cbt1:
Moves the functions for handling item size in react-window list compont to a util-file. It allows us to test rather complex logic without running the whole components.


a-m-dev (approved):

src/pivot-table/components/helpers/__tests__/get-item-size-handler.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+describe("getItemSizeHandler", () => {
+  let list: List;
+
+  beforeEach(() => {
+    list = {};
+  });
+
+  describe("getRowHeightHandler", () => {
+    const cellHeight = 24;
```
a-m-dev, line 13: shouldn't this default to 28px?

PR #275, opened 2023-05-12
feat: pagination (synced with new way of data fetching)

a-m-dev:
pagination feature, including:

- [x] moving page info to supernova level
- [x] introducing `Pagination` component
- [x] current `MAX_COL/MAX_ROW` is set to 15K (fair performance when we have both x and y in 15K cells length)
- [x] tests probably in another pr just to keep this one small **(Draft PR here: https://github.com/qlik-oss/sn-pivot-table/pull/283)**

**UPDATE_#01:** ~~There is an issue regarding the very last page, which does not cut rows and show empty rows until the end but will address it in another pr probably~~ **(Draft PR here: https://github.com/qlik-oss/sn-pivot-table/pull/281)**

**UPDATE_#02:** Closing https://github.com/qlik-oss/sn-pivot-table/pull/257 in favour of this PR

# Pls Note:
The main logic chunk of this PR is about ~300 to ~350 lines of changes, the REST is snapshot test update


src/hooks/use-load-data-pages.ts
```diff
@@ -46,19 +51,18 @@ const useLoadDataPages = (model: Model, layoutService: LayoutService, viewServic
...
+        const fetchArea: EngineAPI.INxPage = {
+          qLeft: qLastExpandedPos ? viewService.gridColumnStartIndex : 0,
+          qTop: qLastExpandedPos ? viewService.gridRowStartIndex : pageInfo.currentPage * pageInfo.rowsPerPage,
+          qWidth: !viewService.gridWidth ? DEFAULT_PAGE_SIZE : viewService.gridWidth,
+          qHeight: !viewService.gridHeight ? DEFAULT_PAGE_SIZE : viewService.gridHeight,
+        };
+        const pivotPages = await (model as EngineAPI.IGenericObject).getHyperCubePivotData(Q_PATH, [fetchArea]);
 
-        setDataPages([...qHyperCube.qPivotDataPages, ...pivotPages]);
+        setDataPages(pivotPages);
```
a-m-dev, line 65: @cbt1  I remember you mentioned why we need to spread previous `qPivotDataPages`, but i dont remember it clearly, but will only including what we receive from websocket cover that case you mentioned before?

cbt1: On `main` we don't really need it. But it does not hurt either, as the more values we have in our initial data the better I'd say.

In this branch my first thought is that it will no longer work. Since when the user have changed page, the initial data pages from the layout are from the first page. That would not match the current page.

src/hooks/use-pagination.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+  } = layoutService;
+
+  const [pageInfo, setPageInfo] = useState<PageInfo>({
+    currentPage: 0,
+    shouldShowPagination: qSize.qcy > size.y,
+    ...getPageMeta(qSize.qcy),
+  });
+
+  useEffect(() => {
```
cbt1, line 35: I think this a good candite for `useOnPropsChange`

cbt1: Oh nevermind its the nebula hook

src/pivot-table/components/PivotTable.tsx
```diff
@@ -68,6 +72,13 @@ export const StickyPivotTable = ({
     }
   }, [layoutService]);
 
+  useLayoutEffect(() => {
+    if (scrollableContainerRef.current && !dataModel.isLoading) {
+      scrollableContainerRef.current.scrollLeft = 0;
+      scrollableContainerRef.current.scrollTop = 0;
+    }
+  }, [pageInfo.currentPage, dataModel.isLoading]);
```
cbt1, line 80: Why `dataModel.isLoading` here?

src/pivot-table/data/extract-left.ts
```diff
@@ -19,6 +20,7 @@ const extractLeftGrid = (
     root: Cell | null,
     parent: Cell | null,
     nodes: EngineAPI.INxPivotDimensionCell[],
+    pgInfo: PageInfo,
```
cbt1, line 23: You don't need to pass pageInfo in as a param here. It's already available in the function scope

src/pivot-table/components/PivotTable.tsx
a-m-dev: previously we needed to wait for that then change the page, probably not needed any more, since the fetching is not happening with dataModel.

src/pivot-table/hooks/__tests__/use-data.test.ts
```diff
@@ -1,41 +1,56 @@
...
+  type AddPageToTopDimensionDataProps,
+} from "../../data/top-dimension-data";
 import useData from "../use-data";
 
+jest.mock("@nebula.js/stardust", () => ({
+  __esModule: true,
+  useState: () => {},
+  useEffect: () => {},
+  useMemo: () => {},
+}));
```
haxxmaxx, line 31: This is really good to know that it works 👍

haxxmaxx (commented):
LGTM!

haxxmaxx (approved):

src/pivot-table/data/extract-left.ts
```diff
@@ -25,10 +26,14 @@ const extractLeftGrid = (
...
-      rowIdx += currRowIdx === 0 ? 0 : 1;
-      const y = qArea.qTop + rowIdx - node.qUp; // Start position + current page position - previous tail size,
-      const cell = createCell(node, parent, root, colIdx, y, isSnapshot);
+    nodes.forEach((node, idx) => {
+      rowIdx += idx === 0 ? 0 : 1;
+      // consider items that might be skipped based on current page
+      const startPosition = qArea.qTop - pageInfo.currentPage * pageInfo.rowsPerPage;
+      // Start position + current page position - previous tail size,
+      const y = Math.max(0, startPosition + rowIdx - node.qUp);
+      const dataY = qArea.qTop + rowIdx - node.qUp;
```
cbt1, line 35: I do think you should call it `y` and `pageY`, so that both `x`and `y` refers to the "data" position. No need to fix in this PR.

src/pivot-table/hooks/__tests__/use-data.test.ts
```diff
@@ -1,41 +1,56 @@
...
+} from "../../data/left-dimension-data";
+import { addPageToMeasureData, createMeasureData, type AddPageToMeasureDataProps } from "../../data/measure-data";
+import {
+  addPageToTopDimensionData,
+  createTopDimensionData,
+  type AddPageToTopDimensionDataProps,
+} from "../../data/top-dimension-data";
 import useData from "../use-data";
 
+jest.mock("@nebula.js/stardust", () => ({
```
cbt1, line 26: This should not be needed anymore since it's no longer a nebula hook?

cbt1 (approved):
Some minor comments. LGTM

src/pivot-table/data/extract-left.ts
a-m-dev: will fix it soon in another PR

PR #278, opened 2023-05-15
feat: theme v2

cbt1:
Consumes the new Theme definition. It enables unique styling across 4 "sections". It also changes the default cell height to "compact", which means 24px instead of 28px.

This PR breaks the Styling panel. So changes in the styling panel will no longer have them same effect. Another PR will follow that updates the styling panel with all options from the new theme.

The reduced of test coverage is the grid components that does have some added logic to them but no unit test exists for them (still covered by rendering tests).

TODO:
- Styling panel
- Total value cell styling
- Total value cell border (divider)
- Color cell background/text by expression
- API Spec

Example:
Skärmavbild 2023-05-16 kl  07 12 20


src/services/theme-types.ts
```diff
@@ -0,0 +1,48 @@
+type Color = string;
```
cbt1, line 1: Fix

nebula.config.js
```diff
@@ -9,4 +9,181 @@ module.exports = {
       "process.env.PACKAGE_VERSION": JSON.stringify(version),
     },
   },
+  serve: {
+    flags: {},
+    keyboardNavigation: false,
+    themes: [
```
cbt1, line 15: Added 3 different themes.
- The "default" has all default values
- "Demo" has a unique value for each property
- "Only transparent borders" show case when cell borders are turned off

a-m-dev (approved):
LGTM ✅

PR #279, opened 2023-05-17
feat: styling panel theme v2

cbt1:
Adds the options from theme v2 to the styling panel

Skärmavbild 2023-05-17 kl  13 14 43


src/types/QIX.ts
```diff
@@ -54,24 +54,24 @@ interface ComponentCellStyling {
 export interface Component {
   key: "general" | "theme";
   rowHeight?: RowHeight;
-  header?: {
```
cbt1, line 57: Should make optional again

src/ext/property-panel/settings/styling-panel/utils/create-color-picker-item.ts
```diff
@@ -0,0 +1,19 @@
+import type { Args, CurrentTheme } from "../../../../../types/QIX";
+
+export const createColorPickerItem = (
+  ref: string,
+  translation: string,
+  themeAccessor: (theme: CurrentTheme) => string
+) => ({
+  ref,
+  translation,
+  // show: true,
```
a-m-dev, line 10: probably forgotten comment as well

src/ext/property-panel/settings/styling-panel/utils/create-font-family-item.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type ThemeAccessor = (currentTheme: CurrentTheme) => string;
+
+interface Props {
+  ref: string;
+  themeAccessor: ThemeAccessor;
+  translator: stardust.Translator;
+}
+
+const DEFAULT_FONT_FAMILIES: string[] = [
```
a-m-dev, line 14: I saw it somewhere in some repo this: https://github.com/qlik-trial/sense-client/blob/b9115ce436c3b8e36e84a1ca9cb4283359a1917c/web/assets/client/utils/theme/theme.js#L342-L348
has been used to access font families, so probably dont we need to hardcode available font families (+ font sizes) here?

a-m-dev (commented):
generally looks good, 
just some minor comments and one concern about how we load the font families and sizes instead of hardcoding them.

a-m-dev (commented):
generally looks good, 
just some minor comments and one concern about how we load the font families and sizes instead of hardcoding them.

src/ext/property-panel/settings/index.ts
```diff
@@ -0,0 +1,74 @@
+// import { TOTAL_MODE_ON } from '../../constants';
```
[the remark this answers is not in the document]
cbt1: Indeed

src/ext/property-panel/settings/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      defaultValue: true,
+      show(properties: EngineAPI.IGenericHyperCubeProperties): boolean {
+        return properties.qHyperCubeDef?.qDimensions?.some((qDim) => qDim.qOtherTotalSpec?.qTotalMode === "TOTAL_EXPR");
+      },
+    },
+    // indentMode: {
+    //   ref: 'qHyperCubeDef.qIndentMode',
+    //   type: 'boolean',
+    //   translation: 'properties.pivot.indentMode',
+    // },
```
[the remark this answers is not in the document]
cbt1: It's been there for since forever, but I'll remove it

src/ext/property-panel/settings/styling-panel/utils/create-font-family-item.ts
cbt1: There is no way to access that. The theme API that is exposed is coming from https://github.com/qlik-oss/nebula.js/blob/3b4b6ece98251d952eff93084bca464e2f1e531d/apis/theme/src/index.js#L32

a-m-dev (approved):

haxxmaxx:
We probably want to use the fontsize and color component when applicable
Screenshot 2023-05-29 at 11 58 26

cbt1:
> We probably want to use the fontsize and color component when applicable Screenshot 2023-05-29 at 11 58 26

Are they exposed? @haxxmaxx

haxxmaxx:
We use them in sn-table [here](https://github.com/qlik-oss/sn-table/blob/crl/break-out-pagination/src/ext/property-panel/settings.js#L23)

cbt1:
> We use them in sn-table [here](https://github.com/qlik-oss/sn-table/blob/crl/break-out-pagination/src/ext/property-panel/settings.js#L23)

I see. The colorpicker component it used. The one for the font-size is just an input field, which I did not like so I wrote a dropdown component instead where you can select a font size. More or less the same as the "General tab" styling

PR #283, opened 2023-05-23
test: pagination

a-m-dev:
here is the promised test PR for [pagination](https://github.com/qlik-oss/sn-pivot-table/pull/275)


src/hooks/__tests__/use-load-data-pages.test.ts
```diff
@@ -0,0 +1,318 @@
...
+
+const usePromiseResult: Promise<void> | undefined = Promise.resolve();
+let usePromiseError: Error | undefined;
+
+const usePromise = (callback: () => void) => {
+  callback();
+  return [usePromiseResult, usePromiseError];
+};
+
+jest.mock("@nebula.js/stardust", () => ({
```
cbt1, line 17: Should look into putting this under `<root-folder>/__mocks__/@nebula.js/stardust.ts` instead so that it is automatically mocked for all tests. Since we'll never be able to run nebula hooks anyway.

src/hooks/__tests__/use-load-data-pages.test.ts
```diff
@@ -0,0 +1,318 @@
...
+      getHyperCubePivotDataMock = jest.fn();
+      model = {
+        getHyperCubePivotData: getHyperCubePivotDataMock,
+      } as unknown as Model;
+    });
+
+    test("should set `qPivotDataPages` from layout if there was any", () => {
+      renderer();
+      expect(useStateSetterMock).toHaveBeenCalledTimes(1);
+      expect(useStateSetterMock).toHaveBeenCalledWith(layoutService.layout.qHyperCube.qPivotDataPages);
```
cbt1, line 181: Can you not verify the return value of the hook instead of looking at the mocks?

src/pivot-table/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const DEFAULT_FONT_SIZE = "12px";
-export const DEFAULT_FONT_FAMILY = "Source Sans Pro, Arial, sans-serif";
+export const DEFAULT_FONT_FAMILY = '"Source Sans Pro", "Arial", "sans-serif"';
```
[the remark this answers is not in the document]
a-m-dev: not sure why i have changed that, will revert it asap

src/hooks/__tests__/use-load-data-pages.test.ts
a-m-dev: not actually, 
Here the challenge is `usePromise`!

we are **setting a state** within it which means **another re-render** of hook -> which causes an **infinite loop** in renderHook and jest kills it with failure error!

Screenshot 2023-05-29 at 16 26 03


So in my solution, I'm just mocking state and checking if the state setter has been called with proper argument, **relying on the guarantee that how `useState` works** makes me sure that if state updates with proper arguments, then I will have a proper output from hook.

Ofcourse better idea is welcome here!

cbt1: Do really understand why it would do that but I'm fine with the solution then 👍

cbt1 (approved):
Dont really have time to do an in-depth review of this but as a whole it looks good to me

PR #285, opened 2023-05-23
build: switched to @scriptappy scope

enell:


haxxmaxx:
Looks good, but what is the difference using the @scriptappy packages?

enell:
@haxxmaxx 
The scriptappy-from-jsdoc package has changed name to @scriptappy/from-jsdoc. We will not get any updates if we stick to the old package

haxxmaxx:
ok then it is a no-brainer

haxxmaxx (approved):

PR #287, opened 2023-05-29
feat: colwidth from property panel

haxxmaxx:
First commit to enable changing the column width using the property panel.
The left grid is pretty simple in terms of calculating the width. It works more or less as a table. That being said, the `auto` and `percentage` setting doesn't really make sense at this point. The total width of the left grid kind of depend on width of the columns in there, so if the column widths depend on the width of the grid, you get loop... (auto tries to fill the width of the grid and percentage means x% of the grid width).

The top grid is inherently more complex. So the setting that is actually decides the width, is the **width of the bottom row, if it is actually showing**. So if the bottom row (leafs) is a dimension, we use those values for those cells. Then for the cells in the row above, they simply get the sum of the leaf widths. In this case the bottom row in the top grid and the entire data grid will have the same width.
Screenshot 2023-08-02 at 14 58 19

For measures the width can differ if you have more than one measure. They are basically independent. The columns in the data grid will correspond to the ones in the leafs of the top grid.
Screenshot 2023-08-02 at 14 56 46

 
Currently the `auto` option does not make sense, if you don't have it for all measures. It will simply calculate the width each cell should have to fill the width. Then some other measure can have a different value, making the total width incorrect. This could possibly be improved.
Screenshot 2023-08-02 at 14 59 39

- [x] type change functions in the property definition, and make them cap of the values you can set for pixels and percentage
- [x] write unit tests
- [x] figure out how we should do auto and percentage for the left grid. Don't think we can take the options away, and if the option does different things depending on where the field is (top or left) would be really confusing for the user. EDIT: We just do the same thing for fit to content for auto. And for percentage we refer to the entire width of the chart. It will be scaled down for now, when it gets to wide. But the left grid will be scrollable eventually and then the 100 makes sense.
- [x] fix borders. should not show the divider for the top grid all the way to the right, and the right side should get a border, when the columns does not fill the entire chart
- [x] add comments and some jsdocs. Not simple to follow all of this without it
- [x] Make the fit to content width the auto min width. Right now it will try to fit as many columns as possible with the min column value

Things that have been scoped out of this PR:
- [ ] Adapt top grid widths to if the expand icon is showing or not. Already have a broken out function to cover that
- [ ] Add a disclaimer in the property panel. Need to show that the setting is not applied if that field is not the bottom one in the top grid
- [ ] look into moving the properties definition to the common repo. Currently there is no difference between the default values or the properties setup
- [ ] Add columnWidth to object-properties
- [ ] handle the specific case when are not on the first page, scroll down and expand a node
- [ ] smarter way of calculating auto, when the other columns don't have auto (basically only applicable when pseudo dimension is at the bottom)


src/pivot-table/components/helpers/get-item-size-handler.ts
```diff
@@ -31,24 +29,20 @@ export const getRowHeightHandler =
...
   (colIndex: number) => {
     const cell = isLastRow ? list[colIndex] : Object.values(list)[colIndex];
-    const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
+    // const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
 
-    if (colIndex === 0 && cell?.x > 0) {
-      return ((cell.leafCount + cell.x) / measureInfoCount) * allMeasuresWidth;
-    }
+    // Dunno when this is true
+    // if (colIndex === 0 && cell?.x > 0) {
```
haxxmaxx, line 38: Never understood when this is true?

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,41 +1,65 @@
...
 
 export const EXPAND_ICON_WIDTH = 30;
-const MIN_COLUMN_WIDTH = 100;
+const MIN_COLUMN_WIDTH = 30;
 const MAX_RATIO_OF_TOTAL_WIDTH = 0.75;
+const DEFAULT_PIXEL_VALUE = 200;
+const DEFAULT_PERCENTAGE_VALUE = 20;
+
+// TODO: refactor and move this to use-measure-text file
+const useMeasureAndEstimate = (styling: HeaderStyling | MeasureContentStyling | DimensionContentStyling) =>
```
haxxmaxx, line 35: This is basically what you always do, so might as well just send in the object will the styling properties. Will look at a refactor, probably in a separate PR

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: It's a bit of a special case but if you are on a different page then the first. Scroll down a bit and expand a node. The "first row" does not exist yet as data has only been fetched for the rows that are visible to the user.

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -49,14 +79,26 @@ export default function useColumnWidth(
...
-      const hasChildNodes = index < qNoOfLeftDims - 1; // -1 as the last column can not be expanded or collapsed
-      const collapseExpandIconSize = hasChildNodes ? EXPAND_ICON_WIDTH : 0;
+      const { qFallbackTitle, qApprMaxGlyphCount, columnWidth } = qDimensionInfo[dimIndex];
+      // percentage
+      if (columnWidth?.type === "pixels") {
+        return (columnWidth.pixels || ColumnWidthValues.PixelsDefault) / rect.width;
+      }
+      // pixels
+      if (columnWidth?.type === "percentage") {
+        // TODO: do we want the percentage value to represent the entire chart or just the 75% that you can occupy with the left side
```
cbt1, line 89: In the future when the left grid can be scrolled horizontally by itself, as in not scrolling the top or measure grids. 100% could mean the entire chart, but the actual visible width of the left grid would be something like 75% of the entire chart. So the user could scroll the remaining 25% in the left grid.

But until then I think it's better have "percentage" value be based on the actual width of the left grid (75% of the entire chart).

src/types/QIX.ts
```diff
@@ -132,12 +131,46 @@ export interface ExtendedDimensionInfo extends EngineAPI.INxDimensionInfo {
...
   };
+  columnWidth: ColumnWidth;
+}
+
+export interface ExtendedMeasureInfo extends EngineAPI.INxMeasureInfo {
+  columnWidth: ColumnWidth;
 }
 
-export interface HyperCube extends EngineAPI.IHyperCube {
+export interface ExtendedHyperCube extends Omit<EngineAPI.IHyperCube, "qDimensionInfo" | "qMeasureInfo"> {
```
haxxmaxx, line 141: chose to call these `ExtendedXyz`. We might want to remove that for shorter names

cbt1:
Currently the default behavior when creating a new chart (in nebula dev tools):

Skärmavbild 2023-09-13 kl  08 13 33

haxxmaxx:
> Currently the default behavior when creating a new chart (in nebula dev tools):
> 
> Skärmavbild 2023-09-13 kl 08 13 33

Hmm yeah when there is no setting, it defualts to the min value, should be the same as auto. good find

haxxmaxx:
I fixed the default values for the right grid when no columnWidth prop exist on the chart properties. Also made the fitToContent the min value for the auto setting.

cbt1:
The default width for a leaf node on the top grid has changed. Example data uses `Product` as top dimension and a single measure.

On `main`:
Skärmavbild 2023-09-13 kl  12 56 45

Now:
Skärmavbild 2023-09-13 kl  12 56 29

a-m-dev:
is this testable only in sense-client currently? and if so, can i ask how you mount your local running pvt in sense-client?

haxxmaxx:
> is this testable only in sense-client currently? and if so, can i ask how you mount your local running pvt in sense-client?

Well you can modify the properties, in the nebula dev env. But that is a bit much work. Easier to test in the client. I can send you a link to a PR

cbt1:
Will you add the new columnWidth properties to the API spec?

haxxmaxx:
> Will you add the new columnWidth properties to the API spec?

Yes, I will write that down. But no need to do it before it is all done right?

cbt1:
> > Will you add the new columnWidth properties to the API spec?
> 
> Yes, I will write that down. But no need to do it before it is all done right?

No, lets just remember to do it 👍

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: FYI this scenario should still remain relevant

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,167 +1,240 @@
...
 
-        return pseudoDimensionWidth / rect.width;
+  /**
+   * The widths of the left columns. Scales the width to fit LEFT_SIDE_MAX_WIDTH_RATIO * rect.width if wider than that
+   */
+  const leftGridColumnWidths = useMemo(() => {
+    const getColumnWidth = (columnWidth: ColumnWidth, fitToContentWidth: number) => {
+      switch (columnWidth?.type) {
+        case ColumnWidthType.Pixels:
+          return columnWidth.pixels || ColumnWidthValues.PixelsDefault;
```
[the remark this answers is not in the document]
haxxmaxx: Good point. We throw an error in sn-table, but that is only for invalid types. Negative numbers will just default to the min value.

src/pivot-table/components/helpers/get-item-size-handler.ts
haxxmaxx: Yeah I figured I would look into it.

haxxmaxx: scoped out of this PR, see description

src/pivot-table/hooks/use-column-width.ts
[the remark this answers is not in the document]
haxxmaxx: done, now should default to default values for any invalid value

a-m-dev:
I saw this in one of rendering test results but compared to the previous version, is this correct behaviour when measures are on top? 
Prev: 
Screenshot 2023-09-21 at 13 03 28

Current:
Screenshot 2023-09-21 at 13 02 22


I mean, for measures on top, it does not even considers the min width of the (measure) labels as before and probably it calculates based on data cells max lenght? I'm just wondering, is it the way we want it to be?

haxxmaxx:
> I mean, for measures on top, it does not even considers the min width of the (measure) labels as before and probably it calculates based on data cells max lenght? I'm just wondering, is it the way we want it to be?

For now, yes. We had this discussions in one of the daily. The _estimated_ width is pretty inaccurate while the _measered_ width is accurate. Now for the case above, before my changes, the width is based on the measured title width and the estimated width of the data. Since the title is much wider, it will adapt to that. In this particular case, it was better before. But for the sake of consistency, we use a fixed min value as default (the auto or no setting) for both dimensions and measure.

If we do it the other way around, use the measure/estimated width for min width, you get the case that Bengtsson posted above. For dimensions, when the dimension values are really long, you can get really wide column, that shows almost no data. So we would rather avoid that. A solution would be a smarter way of estimating the width. But until then it can produce some very inaccurate results.

TL;DR Yes, we want the min value to be consistent and simple, at least until we can estimate text length for dimension values better

src/ext/property-panel/data.ts
```diff
@@ -26,6 +28,69 @@ function isTotalsVisible(itemData: EngineAPI.IHyperCubeDimensionDef, _: unknown,
...
+        translation: "Object.Table.Column.Pixels",
+      },
+      {
+        value: ColumnWidthType.Percentage,
+        translation: "Object.Table.Column.Percentage",
+      },
+    ],
+    defaultValue: ColumnWidthType.Auto,
+  },
+  sizePixels: {
```
cbt1, line 58: I'm getting some unexpected result (for me) when switching to pixel based size via PP. The default value is displayed in the PP but the layout is not actually updated with the value. Only the `type` property exist on `qMeasureInfo.columnWidth`.

I'd say it works fine because ther is likely some fallback logic when resolving the pixel based column width. Still worth noting I think.

src/ext/property-panel/data.ts
```diff
@@ -26,6 +28,69 @@ function isTotalsVisible(itemData: EngineAPI.IHyperCubeDimensionDef, _: unknown,
...
+      {
+        value: ColumnWidthType.Auto,
+        translation: "Common.Auto",
+      },
+      {
+        value: ColumnWidthType.FitToContent,
+        translation: "Object.Table.Column.FitToContent",
+      },
+      {
+        value: ColumnWidthType.Pixels,
```
cbt1, line 48: Actually that comment about pixel type should be here and I guess I kind of make a bit more sense when switching type. This does not set the actual pixel value on the properties.

src/pivot-table/components/helpers/__tests__/get-item-size-handler.test.ts
```diff
@@ -145,41 +114,32 @@ describe("getItemSizeHandler", () => {
...
-        layoutService,
-        getMeasureInfoWidth,
-        allMeasuresWidth,
-      });
-
-      expect(handler(index)).toEqual(
-        (allMeasuresWidth * (leafCount + x)) / layoutService.layout.qHyperCube.qMeasureInfo.length,
-      );
-    });
+    // test("should return a size for cell when first column does not exist in list", () => {
```
cbt1, line 121: Why is this commented?

src/pivot-table/components/helpers/get-item-size-handler.ts
```diff
@@ -31,24 +29,20 @@ export const getRowHeightHandler =
...
+  ({ list, isLastRow, getRightGridColumnWidth }: ColumnWidthHandlerProps): ItemSizeHandler =>
   (colIndex: number) => {
     const cell = isLastRow ? list[colIndex] : Object.values(list)[colIndex];
-    const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
+    // const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
 
-    if (colIndex === 0 && cell?.x > 0) {
-      return ((cell.leafCount + cell.x) / measureInfoCount) * allMeasuresWidth;
-    }
+    // Dunno when this is true
```
cbt1, line 37: Can you add this:
> It's a bit of a special case but if you are on a different page then the first. Scroll down a bit and expand a node. The "first row" does not exist yet as data has only been fetched for the rows that are visible to the user.

As a comment here instead so that it does not get lost? I good to enough the specific scenario this logic is covering

src/types/QIX.ts
cbt1: Do you actually need to omit the dim and measure info here? Since they get overridden anyway

src/types/QIX.ts
```diff
@@ -132,14 +131,48 @@ export interface ExtendedDimensionInfo extends EngineAPI.INxDimensionInfo {
...
+// types for properties
+interface ExtendedInlineDimensionDef extends EngineAPI.INxInlineDimensionDef {
+  columnWidth: ColumnWidth;
+}
+
+interface ExtendedInlineMeasureDef extends EngineAPI.INxInlineMeasureDef {
+  columnWidth: ColumnWidth;
+}
+
+export interface DimensionOrMeasure {
```
cbt1, line 176: Maybe rename to indicate that it's properties. Ex: `DimensionOrMeasureDef` or `DimensionOrMeasureProps`

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: Doing some testing I found a common case where the scroll position is lost when expanding a node. Just scroll a bit to a node and expand it. Doing that seem to cause the scroll position to get lost as it's no longer positioned at the expanded node.

This scenario works just fine on main. I'll do some more digging on this.

cbt1: Seem to only happen when doing it on the top grid

src/ext/property-panel/data.ts
haxxmaxx: yeah good point, I'll look at that. Maybe we want to set the value for pixels when changing the type to pixels

src/types/QIX.ts
haxxmaxx: good point

src/pivot-table/components/helpers/get-item-size-handler.ts
haxxmaxx: haha yep the current one is pretty bad

src/pivot-table/components/helpers/__tests__/get-item-size-handler.test.ts
haxxmaxx: It is the corner case that I commented out. I thought I would keep the test since, for when I fix that case

src/ext/property-panel/data.ts
haxxmaxx: will scope it out of this PR in any case. I can do it as I add the disclaimer

src/pivot-table/components/helpers/get-item-size-handler.ts
haxxmaxx: Do you think I need to fix it in this one? I can look at it the first thing in the next PR

src/types/QIX.ts
haxxmaxx: Don't know I followed the pattern from sn-table

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: Most likely not, sure yet. Need to figure out when and why it happens. If it's a likely scenario to trigger or not

src/types/QIX.ts
haxxmaxx: quick google and seems like you get a type error without it, but it does not complain when I try to remove it

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: Something is certianly broken as it fails to render cells in this case. I just scroll right (more than 50 columns) and expanded a node

Skärmavbild 2023-09-22 kl  10 26 57

src/pivot-table/components/helpers/get-item-size-handler.ts
```diff
@@ -31,24 +29,22 @@ export const getRowHeightHandler =
...
+    // const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
 
-    if (colIndex === 0 && cell?.x > 0) {
-      return ((cell.leafCount + cell.x) / measureInfoCount) * allMeasuresWidth;
-    }
+    // TODO: This a bit of a special case but if you are on a different page then the first.
+    // Scroll down a bit and expand a node. The "first row" does not exist yet as data has only been
+    // fetched for the rows that are visible to the user.
+    // if (colIndex === 0 && cell?.x > 0) {
+    //   return ((cell.leafCount + cell.x) / measureInfoCount) * allMeasuresWidth;
```
cbt1, line 41: Changing this to `return (cell.leafCount + cell.x) * getRightGridColumnWidth();` seem to fix the issue where it fails to render cells at least

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,167 +1,245 @@
...
 }
 
 export const EXPAND_ICON_WIDTH = 30;
-const MIN_COLUMN_WIDTH = 100;
-const MAX_RATIO_OF_TOTAL_WIDTH = 0.75;
+const LEFT_GRID_MAX_WIDTH_RATIO = 0.75;
+
+export enum ColumnWidthValues {
+  PixelsMin = 30,
+  PixelsMax = 7680,
```
a-m-dev, line 29: why we should allow this much size on a col? is there any reason why that number?

haxxmaxx: It's the width of an 8k display. Used the same as for sn-table. My reasoning was that if you for some reason wanted a single column table, that takes up the entire screen, this is the width you need. It is pretty arbitrary, could lower it

src/pivot-table/components/helpers/get-item-size-handler.ts
haxxmaxx: great, I will test a bit with that

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: > Doing some testing I found a common case where the scroll position is lost when expanding a node. Just scroll a bit to a node and expand it. Doing that seem to cause the scroll position to get lost as it's no longer positioned at the expanded node.

This seem to be because the width of the columns changes as the data changes when you expand/collapse. It's on main as well. So that should be fine to deliver with.

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,167 +1,245 @@
...
+  PixelsDefault = 200,
+  PercentageMin = 1,
+  PercentageMax = 100,
+  PercentageDefault = 20,
+  AutoMin = 80,
+}
+
+const getValidValue = (value: number | undefined, defaultValue: number) =>
+  !!value && typeof value === "number" && !Number.isNaN(value) ? value : defaultValue;
```
a-m-dev, line 38:
```suggestion
  !!value && !Number.isNaN(value) ? value : defaultValue;
```

if value is not falsy -> `!!value` will take care of `| undefined` case and there after the value argument will be treated as number. 
So i presume you dont need to check `typeof` it? unless typescript complains about something else

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,167 +1,245 @@
...
+    const widths = visibleLeftDimensionInfo.map((qDimensionInfo, index) => {
+      let width;
+
+      if (qDimensionInfo === PSEUDO_DIMENSION_INDEX) {
+        // Use the max width of all measures
+        width = Math.max(
+          ...qMeasureInfo.map(({ qFallbackTitle, columnWidth }) => {
+            const fitToContentWidth = measureTextForContent(qFallbackTitle);
+            return getColumnWidth(columnWidth, fitToContentWidth);
+          }),
```
a-m-dev, line 99: I think this can be externalised from it's outer map which happens on line 90 `visibleLeftDimensionInfo.map(...)`.
currently it's a small O(n*2) operation happens, I expect it's not that much heavy operation for all executions since it's only happens for pseudo dims. but it's good to externalise it since what it requires is only `qMeasureInfo` which `visibleLeftDimensionInfo` is not coupled with it at all.

a-m-dev (commented):
Generally Looks great

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,167 +1,245 @@
...
-      Array.from({ length: size.x }, () => null).reduce(
-        (width, _, index) => width + getDataColumnWidth(index),
-        leftGridWidth,
-      ),
-    [getDataColumnWidth, leftGridWidth, size.x],
+  /**
+   * Gets the width of a right grid column. This is always based on the leaf width(s)
+   */
+  const getRightGridColumnWidth = useCallback(
+    (index?: number) => (topGridLeavesIsPseudo && index !== undefined ? memoizedGetLeafWidth(index) : averageLeafWidth),
```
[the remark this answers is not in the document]
a-m-dev: oops, i forgot 0 index 😵, let's keep it as is

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: Because strings that are only number return false from isNaN, e.g. `'1'`

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: Well, there can only be one pseudo dimension in `visibleLeftDimensionInfo`, so `qMeasureInfo.map` can only run once

src/pivot-table/hooks/use-column-width.ts
a-m-dev: you should not have numbers as strings because you are getting value as number (`value: number`) in arguments

haxxmaxx: Yeah if the user follows the typing. I think you can set it to a string, then we want to cover that. But maybe it is overkill

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -1,167 +1,245 @@
...
+          }),
+        );
+      } else {
+        const { qFallbackTitle, qApprMaxGlyphCount, columnWidth } = qDimensionInfo;
+        const fitToContentWidth = Math.max(
+          measureTextForHeader(qFallbackTitle),
+          estimateWidthForRowContent(qApprMaxGlyphCount) + getCollapseExpandIconSize(index),
+        );
+
+        width = getColumnWidth(columnWidth, fitToContentWidth);
```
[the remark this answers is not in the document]
haxxmaxx: Good point. I followed the same pattern as in the right side width. The thing is that there really is no specific auto setting, where the columns try to fit the width of the chart. So value becomes the same for fitToContent, auto and default. Maybe we will introduce auto in the future, but hard to see how that logic would work. But I do like that we extract all the logic into one place. With your suggestion, we would have `getColumnWidth(columnWidth) ?? fitToContentWidt` in two places, and would have to change in two places, if the logic changes in the future

a-m-dev (approved):
There are some minor issues to address before merging:
1. The commented code about possible edge case
2. [MINOR]: measure label ... not showing 
https://github.com/qlik-oss/sn-pivot-table/assets/29652890/9fe3db5b-a237-48ab-8f3d-617100bdfa3e

Other than that LGTM!

**UPDATE 01:**
I forgot to attach the screenshot that i was recorded for number 2(bunch of things happened at the same time 🥴):

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/1abf3123-0ab3-4a0f-a179-588f9b13c975

cbt1 (approved):
LGTM

cbt1:
> There are some minor issues to address before merging:
> 
>     1. The commented code about possible edge case
> 
>     2. [MINOR]: measure label ... not showing
>        https://github.com/qlik-oss/sn-pivot-table/assets/29652890/9fe3db5b-a237-48ab-8f3d-617100bdfa3e
> 
> 
> Other than that LGTM!

Regarding nbr 2. I think the width is the expected? But would have expected there to be ellipsing?

haxxmaxx:
> Regarding nbr 2. I think the width is the expected? But would have expected there to be ellipsing?

Yes it is the ellipses missing. It is there if I open the test app, so an actual issue

PR #288, opened 2023-05-29
fix: remove min and max width for svg element

enell:
Chrome and Firefox calculates fit-content differently than Safari. The
minSize isn't necessary for the svg component.

Refs: VIZ-2584


src/pivot-table/components/icons/Minus.tsx
```diff
@@ -18,7 +17,6 @@ const MinusIcon = ({ color, opacity, testid, onClick }: IconProps): JSX.Element
     xmlns="http://www.w3.org/2000/svg"
     onClick={onClick}
     data-testid={testid}
-    style={{ minWidth: "fit-content", minHeight: "fit-content" }}
```
cbt1, line 21: I recall seeing an issue where the icons would not always render with the same size. That is why the styling was added. I'll see if I can reproduce that in this branch.

cbt1: This does seem to re-introduce the issue seen in #246 

At least for me in Firefox

Skärmavbild 2023-05-30 kl  07 58 47

enell: I'll investigate more. It is specifically fit-content that isn't working in webkit

haxxmaxx (approved):
LGTM

PR #302, opened 2023-06-28
fix: selection confirm should reset page to the first one

a-m-dev:
confirming selection is now resetting page to 0

### DIFF
--- a/src/pivot-table/hooks/__tests__/use-selections-model.test.ts
+++ b/src/pivot-table/hooks/__tests__/use-selections-model.test.ts
@@ -6,6 +6,7 @@ import useSelectionsModel from "../use-selections-model";
 describe("useSelectionsModel", () => {
   let selections: ExtendedSelections;
   let updatePageInfo: jest.MockedFunction<(args: Partial<PageInfo>) => void>;
+  let callbacks: Record<string, () => void>;
 
   beforeEach(() => {
     selections = {
@@ -16,9 +17,12 @@ describe("useSelectionsModel", () => {
       begin: () => Promise.resolve(),
     } as unknown as ExtendedSelections;
 
+    callbacks = {};
     updatePageInfo = jest.fn();
 
-    jest.spyOn(selections, "on").mockImplementation((_evt, callback) => callback());
+    jest.spyOn(selections, "on").mockImplementation((_evt, cb) => {
+      callbacks[_evt] = cb;
+    });
     jest.spyOn(selections, "removeListener");
     jest.spyOn(selections, "select");
     jest.spyOn(selections, "isActive");
@@ -26,6 +30,7 @@ describe("useSelectionsModel", () => {
   });
 
   afterEach(() => {
+    callbacks = {};
     jest.resetAllMocks();
   });
 
@@ -50,6 +55,7 @@ describe("useSelectionsModel", () => {
     const { result } = renderHook(() => useSelectionsModel(selections, updatePageInfo));
 
     await act(async () => {
+      Object.entries(callbacks).forEach(([, cb]) => cb());
       await result.current.select(NxSelectionCellType.NX_CELL_TOP, 0, 1)();
       await result.current.select(NxSelectionCellType.NX_CELL_TOP, 0, 2)();
     });
@@ -67,6 +73,7 @@ describe("useSelectionsModel", () => {
     const { result } = renderHook(() => useSelectionsModel(selections, updatePageInfo));
 
     await act(async () => {
+      Object.entries(callbacks).forEach(([, cb]) => cb());
       await result.current.select(NxSelectionCellType.NX_CELL_TOP, 0, 1)();
     });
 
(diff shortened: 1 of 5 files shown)


src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -16,25 +16,30 @@ export interface SelectedPivotCell {
...
+): SelectionModel {
   const isActive = selections.isActive();
   const [selected, setSelected] = useState<SelectedPivotCell[]>([]);
 
   useEffect(() => {
-    const clearSelections = () => setSelected([]);
-    const confirmOrCancelSelections = () => setSelected([]);
+    const clearSelections = () => {
+      setSelected([]);
+      updatePageInfo({ currentPage: 0 });
```
haxxmaxx, line 29: I'm not sure this is the correct place to change the page. Some of these events do not update the data, then you shouldn't need to change page. Is it possible to look at if the data has actually changed and then reset the page?
We do that in sn-table

a-m-dev: There are 4 events that this `clearSelections` will run: 
1. deactivated
2. canceled
3. confirmed
4. cleared

I'll reset the page only in **confirmed** event then which we are sure that user has to see page 0.

Also in this level of component tree, we have access to the unprocessed data (data before separation + transformation into left, top and data grid), so Im not sure about checking/comparing data in that level.

haxxmaxx: let me test it a bit, that might be good enough

src/pivot-table/hooks/__tests__/use-selections-model.test.ts
```diff
@@ -53,41 +56,46 @@ describe("useSelectionsModel", () => {
...
     });
 
     await waitFor(() => expect(result.current.isSelected(NxSelectionCellType.NX_CELL_TOP, 0, 1)).toBeTruthy());
+    await waitFor(() => expect(updatePageInfo).toHaveBeenCalledWith({ currentPage: 0 }));
 
     await act(async () => {
       await result.current.select(NxSelectionCellType.NX_CELL_TOP, 0, 1)();
     });
 
+    await waitFor(() => expect(updatePageInfo).toHaveBeenCalledWith({ currentPage: 0 }));
```
haxxmaxx, line 93: are the selections being confirmed here? Otherwise upadatePageInfo sholudn't be called, right?

a-m-dev: You are right, selections are not confirmed here. In fact, we cannot mock the confirm behaviour in all of these test cases here(or at least i have no idea). And to answer for the question of why the test case is still working currently: this happens because of event registration in useEffect where we register all 4 events. By running the hook, this effect triggers and seems like those functions running once in that step! 

So for now it's better to only check it with something like `toHavebeenCalledTimes(1)` to make sure it's registered correctly.

src/pivot-table/hooks/__tests__/use-selections-model.test.ts
```diff
@@ -15,19 +17,25 @@ describe("useSelectionsModel", () => {
       begin: () => Promise.resolve(),
     } as unknown as ExtendedSelections;
 
-    jest.spyOn(selections, "on");
+    callbacks = {};
+    updatePageInfo = jest.fn();
+
+    jest.spyOn(selections, "on").mockImplementation((_evt, cb) => {
```
haxxmaxx, line 23: I guess there should be no _ before evt since you are using it now

src/pivot-table/hooks/__tests__/use-selections-model.test.ts
```diff
@@ -44,39 +52,66 @@ describe("useSelectionsModel", () => {
   });
 
   test("should select cell and call begin selection", async () => {
-    const { result } = renderHook(() => useSelectionsModel(selections));
+    const { result } = renderHook(() => useSelectionsModel(selections, updatePageInfo));
 
     await act(async () => {
+      Object.entries(callbacks).forEach(([, cb]) => cb());
```
haxxmaxx, line 58: I still don't think this fits in this test. It is only testing what happens when you select a cell and that begin should be called. It does not test the listeners that you attach with `on`. It makes more sense to me if we have a seperate test, looping over all callbacks.

src/pivot-table/hooks/__tests__/use-selections-model.test.ts
```diff
@@ -44,39 +52,66 @@ describe("useSelectionsModel", () => {
...
+      await result.current.select(NxSelectionCellType.NX_CELL_TOP, 0, 1)();
+    });
+
+    act(() => {
+      // trigger callback only if eventKey is "confirmed"
+      Object.entries(callbacks).forEach(([_evtKey, cb]) => {
+        if (_evtKey === "confirmed") cb();
+      });
+    });
```
haxxmaxx, line 100: And here I don't see a point of first calling all callbacks, and then seperately running the confirm callback. I'd say the idea of the test is good, just that running confirm once would be more accurate to what happens in the real implementation

a-m-dev: Correct, since we have all event types in hashmap now, we can directly call them using their key such as `callbacks.confirmed()` and avoid loop and calling other event callbacks

src/pivot-table/hooks/__tests__/use-selections-model.test.ts
```diff
@@ -59,24 +67,44 @@ describe("useSelectionsModel", () => {
 
   test("should select cell and not call begin selection when already active", async () => {
     selections.isActive = () => true;
-    const { result } = renderHook(() => useSelectionsModel(selections));
+    const { result } = renderHook(() => useSelectionsModel(selections, updatePageInfo));
 
     await act(async () => {
+      Object.entries(callbacks).forEach(([, cb]) => cb());
```
haxxmaxx, line 73: Same here, doesn't make sense to run the callbacks

src/pivot-table/hooks/__tests__/use-selections-model.test.ts
```diff
@@ -59,24 +67,44 @@ describe("useSelectionsModel", () => {
...
+    await act(async () => {
+      callbacks.confirmed();
+      await result.current.select(NxSelectionCellType.NX_CELL_TOP, 0, 1)();
+    });
+
+    // trigger callback only if eventKey is "confirmed"
+    act(() => callbacks.confirmed());
+
+    // make sure it's been called 2 times
+    await waitFor(() => expect(updatePageInfo).toHaveBeenCalledTimes(2));
```
haxxmaxx, line 96: Still, why call confirmed twice?

a-m-dev: because we want to simulate the confirm action happens after select call. 

we know that it happens useEffect once. 
but here, manually we do it one more time to say that after select call in line 89 confirm happens. 

then it's just a check to making sure that we call `updatePageInfo` 2 times  ( which is within the `clearSelectionAndResetPage` function that registered to confirmed event)

haxxmaxx: No it is not called twice. The use effect calls `selection.on` that passes the callback to the listener, it does not run the callback. The callback runs when there is a confirm event, which we can simulate in the test now by `callbacks.confirmed()`.

haxxmaxx (approved):

PR #303, opened 2023-06-28
fix: missing page

a-m-dev:
if you expand rows and consecutively end up having more pages than when you opened the app, collapsing a row in last page may cause to land in a empty page (still being the last page even though having less than that). this happens because of how engine works on expand and collapsing the rows. 

### solution:
in case of collapsing and page loss, land user in the actual last page. 

Here is the demo for fix:

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/67f53c68-3b06-417f-bc7d-608364753c6c

### DIFF
--- a/src/hooks/use-pagination.ts
+++ b/src/hooks/use-pagination.ts
@@ -42,6 +42,16 @@ const usePagination: UsePagination = (layoutService) => {
     }));
   }, [layoutService.layout.qHyperCube.qSize.qcy, setPageInfo]);
 
+  useEffect(() => {
+    const { currentPage, totalPages } = pageInfo;
+
+    // currPage is base 0 and totalPages always includes remainder rows in last page
+    // so we need to consider both of them for prevent landing in missing page
+    if (currentPage + 1 > totalPages) {
+      setPageInfo({ ...pageInfo, currentPage: totalPages - 1 });
+    }
+  }, [pageInfo]);
+
   const updatePageInfo: UpdatePageInfo = (args) => setPageInfo({ ...pageInfo, ...args });
 
   return {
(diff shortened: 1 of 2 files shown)


src/hooks/use-pagination.ts
```diff
@@ -42,6 +42,16 @@ const usePagination: UsePagination = (layoutService) => {
...
   }, [layoutService.layout.qHyperCube.qSize.qcy, setPageInfo]);
 
+  useEffect(() => {
+    const { currentPage, totalPages } = pageInfo;
+
+    // currPage is base 0 and totalPages always includes remainder rows in last page
+    // so we need to consider both of them for prevent landing in missing page
+    if (currentPage + 1 > totalPages) {
+      setPageInfo({ ...pageInfo, currentPage: totalPages - 1 });
+    }
```
haxxmaxx, line 52: You could also go back to the first page. But I think this is better. Would it make sense to do this for selections as well? Would be nice to simplify and have one solution. We are breaking the pattern that sn-table follows though. Whenever the number of rows change, we go back to the first page. What do you think?

a-m-dev: whenever number of rows change, "in case of selection" it definitely makes sense to go to the first page (which both table and PV is respecting, after the other PR is being merged)

but what makes PT different than table is **the expansion of a row**, because if that is the case (row changes -> always go to the first page) expanding a row (in page 3 for example) would land user in first page, which we dont want that to happen.

This check is only addressing the collapse scenario where user can still remains in a page that is no more existing.

haxxmaxx: Agree, I think this is the most natural way to do it

haxxmaxx (approved):

PR #304, opened 2023-07-03
refactor: `dataY` and `y`

a-m-dev:
with this change, in a Cell object:
`x` reflects the x index of row in entire dataset
`y` reflects the y index of row in entire dataset
`pageX` reflects x index of row in page (currently its value is same as `x` -> because we don't have horizontal pagination)
`pageY` reflects y index of row in page 

Screenshot 2023-07-20 at 10 27 13

### DIFF
--- a/src/pivot-table/components/cells/ListCellFactory.tsx
+++ b/src/pivot-table/components/cells/ListCellFactory.tsx
@@ -72,7 +72,7 @@ const ListCellFactory = ({ index, style, data }: ListCallbackProps): JSX.Element
     <DimensionCell
       cell={cell}
       data={data}
-      rowIndex={cell.dataY}
+      rowIndex={cell.pageY}
       colIndex={cell.x}
       style={style}
       isLeftColumn={isLeftColumn}
--- a/src/pivot-table/components/cells/ListCellFactory.tsx
+++ b/src/pivot-table/components/cells/ListCellFactory.tsx
@@ -72,7 +72,7 @@ const ListCellFactory = ({ index, style, data }: ListCallbackProps): JSX.Element
     <DimensionCell
       cell={cell}
       data={data}
-      rowIndex={cell.y}
+      rowIndex={cell.dataY}
       colIndex={cell.x}
       style={style}
       isLeftColumn={isLeftColumn}
--- a/src/pivot-table/data/helpers/assign-distance-to-next-cell.ts
+++ b/src/pivot-table/data/helpers/assign-distance-to-next-cell.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-param-reassign */
 import type { Grid, PageInfo, Point } from "../../../types/types";
 
-const assignDistanceToNextCell = (data: Grid, dirKey: "x" | "dataY", size: Point, pageInfo?: PageInfo) => {
+const assignDistanceToNextCell = (data: Grid, dirKey: "x" | "pageY", size: Point, pageInfo?: PageInfo) => {
   data.slice(0, -1).forEach((list) => {
     Object.values(list).forEach((cell, index, cells) => {
       const nextSibling = cells[index + 1];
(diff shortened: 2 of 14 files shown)


src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -11,8 +11,8 @@ const createCell = (
 ): Cell => ({
   ref: node,
   x,
-  y, // position of cell in page
-  dataY, // position of cell in dataset
+  y, // Y index of cell in dataset
+  dataY, // Y index of cell in page
```
haxxmaxx, line 15: I think the naming we better before actually. to me data refers to the entire data set -> dataY should be the y-index of the dataset

a-m-dev: kinda both solution makes sense: 
**Option A, as you saying:**
**y** -> reflects rows Y in page
**dataY** -> reflects rows Y in entire dataset

**Option B, the current situation in this PR:**
**y** -> reflects rows Y in dataset
**dataY** -> reflects rows Y in page

So as we discussed it before with Christofer, our choice was like option B and we created [a task for it](https://jira.qlikdev.com/browse/VIZ-2606) to remind us about. 

but, maybe we should just keep this pr open for further investigation and when Christofer is back maybe!

cbt1: Only "y" needs to separate between the dataset value and the page value. So "x" always refers to the dataset. IMO it would make sense to have "y" refer to the dataset and "pageY" to the page value.

a-m-dev: Great to hear from you Christofer, Thanks!


so if `y` refers to dataset's Y (as of this pr does), then the only change this pr needs is the renaming of `dataY` to `pageY`


### My Concern number 1
i would like to draw your attention to [this part of change](https://github.com/qlik-oss/sn-pivot-table/blob/5acce3665dc6cd2a8279ee291a6593838a23c336/src/pivot-table/data/helpers/assign-distance-to-next-cell.ts#L23) which is in consequence of this name change. it makes the logic it a bit dirty!

previously since y was referring to the actual y per page, it was so convenient to just use it as direction as well (when we were freely passing it to the `size` obj) but now, we need a check for it, since what we are looking for is stored in `dataY` key.


### My Concern number 2
is about the over all x and y values in the cell, because with this change: 
1. `x` will refer to x in dataset
2. `dataY` will refer to y in dataset
3. and finally `y` will refer to pageY

This makes these key names totally inconsistent in cell object. So the solution here would be also change the `x` to something like `dataX` but then the `y` will be left alone in that case and it will not have any counterparts (like another `cell.x` key that refers to pageX for example)

**The general problem** we have here is that we are missing some important keys to keep track of each rows from 2 point of views:
1. **Dataset point of view** (x and y keys)
2. **Page point of view** (pageX and pageY keys)
we need to implement 👆 IMO.

haxxmaxx: >     1. **Dataset point of view** (x and y keys)
> 
>     2. **Page point of view** (pageX and pageY keys)
>        we need to implement 👆 IMO.

I think this is the way to go. This is what we eventually ended up with in sn-table. AFAIK `pageX` and `x` will be the same, but that is fine by me. If we introduce some kind of pagination horizontally we have a variable for it

src/pivot-table/data/helpers/assign-distance-to-next-cell.ts
```diff
@@ -1,26 +1,27 @@
 /* eslint-disable no-param-reassign */
 import type { Grid, PageInfo, Point } from "../../../types/types";
 
-const assignDistanceToNextCell = (data: Grid, direction: "x" | "y", size: Point, pageInfo?: PageInfo) => {
+const assignDistanceToNextCell = (data: Grid, dirKey: "x" | "pageY", size: Point, pageInfo?: PageInfo) => {
```
haxxmaxx, line 4: I guess you can use pageX here as well for consistency?

src/pivot-table/components/cells/ListCellFactory.tsx
```diff
@@ -72,7 +72,7 @@ const ListCellFactory = ({ index, style, data }: ListCallbackProps): JSX.Element
     <DimensionCell
       cell={cell}
       data={data}
-      rowIndex={cell.y}
+      rowIndex={cell.pageY}
```
cbt1, line 75: Just noticed, for another refactor PR we could remove `rowIndex` and `colIndex` props. The data is already available in the `cell` prop. That would make it more obvious what kind of positional data it is in the DimensionCell component.

cbt1 (approved):
LGTM

src/pivot-table/components/cells/ListCellFactory.tsx
a-m-dev: I'll create a refactor ticket for it 👍

PR #318, opened 2023-08-09
fix: locked dimension styling

cbt1:
Fixes an issue where the background styling for a locked dimension was not rendered when the pseudo dimension was not at either the last row or column.

Also included:
* A minor performance improvement as the logic for checking if a cell is locked by the dimension is no longer run at every render. Instead only when new layout is received.
* `background` and `backgroundImage` style properties did not work well together in the DimensionCell component. Now only `background` is used.

TODO for another PR:
* Improve the support for custom background colors. Currently the background for a locked dimension does not take the background set by theme or styling panel into account.
* Remove `dimensionInfoIndexMap` as it should no longer be needed.

Before:
Skärmavbild 2023-08-09 kl  12 32 27

After:
Skärmavbild 2023-08-09 kl  12 28 03

### DIFF
--- a/src/pivot-table/components/cells/DimensionCell.tsx
+++ b/src/pivot-table/components/cells/DimensionCell.tsx
@@ -99,9 +99,7 @@ const DimensionCell = ({
   const { select, isSelected, isActive, isLocked } = useSelectionsContext();
   const isNull = qType === NxDimCellType.NX_DIM_CELL_NULL;
   const selectionCellType = isLeftColumn ? NxSelectionCellType.NX_CELL_LEFT : NxSelectionCellType.NX_CELL_TOP;
-  const isCellLocked =
-    isLocked(selectionCellType, cell.y, colIndex) ||
-    layoutService.isDimensionLocked(selectionCellType, cell.y, colIndex);
+  const isCellLocked = isLocked(selectionCellType, cell.y, colIndex) || cell.isLockedByDimension;
   const isNonSelectableCell = isCellLocked || qType === NxDimCellType.NX_DIM_CELL_EMPTY || constraints.active || isNull;
   const isCellSelected = isSelected(selectionCellType, cell.y, colIndex);
   const resolvedTextStyle = getTextStyle({
--- a/src/services/utils/get-sorted-dimension-info.ts
+++ b/src/services/utils/get-sorted-dimension-info.ts
@@ -1,0 +1,26 @@
+import { PSEUDO_DIMENSION_INDEX } from "../../constants";
+import type { ExtendedDimensionInfo, PivotLayout, PseudoDimension } from "../../types/QIX";
+
+const filterPseudoAndVisibleDimensions = (qDim: ExtendedDimensionInfo | -1) =>
+  qDim === PSEUDO_DIMENSION_INDEX || qDim.qCardinalities.qHypercubeCardinal > 0;
+
+const getSortedDimensionInfo = (layout: PivotLayout) => {
+  const { qHyperCube } = layout;
+  const { qNoOfLeftDims, qEffectiveInterColumnSortOrder, qDimensionInfo } = qHyperCube;
+
+  const sortedDimensionInfo = qEffectiveInterColumnSortOrder.map((index) => {
+    if (index === PSEUDO_DIMENSION_INDEX) return PSEUDO_DIMENSION_INDEX;
+
+    return qDimensionInfo[index];
+  }) as (ExtendedDimensionInfo | PseudoDimension)[];
+
+  const sortedLeftDimensionInfo = sortedDimensionInfo.slice(0, qNoOfLeftDims).filter(filterPseudoAndVisibleDimensions);
+  const sortedTopDimensionInfo = sortedDimensionInfo.slice(qNoOfLeftDims).filter(filterPseudoAndVisibleDimensions);
+
+  return {
+    sortedLeftDimensionInfo,
+    sortedTopDimensionInfo,
+  };
+};
+
+export default getSortedDimensionInfo;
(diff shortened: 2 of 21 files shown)


src/services/utils/get-sorted-dimension-info.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const filterPseudoAndVisibleDimensions = (qDim: ExtendedDimensionInfo | -1) =>
+  qDim === PSEUDO_DIMENSION_INDEX || qDim.qCardinalities.qHypercubeCardinal > 0;
+
+const getSortedDimensionInfo = (layout: PivotLayout) => {
+  const { qHyperCube } = layout;
+  const { qNoOfLeftDims, qEffectiveInterColumnSortOrder, qDimensionInfo } = qHyperCube;
+
+  const sortedDimensionInfo = qEffectiveInterColumnSortOrder.map((index) => {
+    if (index === PSEUDO_DIMENSION_INDEX) return PSEUDO_DIMENSION_INDEX;
```
a-m-dev, line 12: I think it would be better to take a solution like defining dim type in `ExtendedDimensionInfo` and then the output of this function will be only `ExtendedDimensionInfo[]` which covers a type of dim called Pseudo. 

That way we could do what ever we want and won't need to check for type of this object where we use it (example in create-cell module in `isLockedByDimension` key)

+ also we can always safely perform higher order functions on it's return value (`sortedDimensionInfo`) without worrying about the type of items in array.

src/services/utils/get-sorted-dimension-info.ts
```diff
@@ -0,0 +1,26 @@
+import { PSEUDO_DIMENSION_INDEX } from "../../constants";
+import type { ExtendedDimensionInfo, PivotLayout, PseudoDimension } from "../../types/QIX";
+
+const filterPseudoAndVisibleDimensions = (qDim: ExtendedDimensionInfo | -1) =>
+  qDim === PSEUDO_DIMENSION_INDEX || qDim.qCardinalities.qHypercubeCardinal > 0;
```
a-m-dev, line 5: just a question out of curiosity:  
Is `qDim.qCardinalities.qHypercubeCardinal > 0` the way we detect a normal dimension?
On the flip side, could it be any other dimension type rather than normal and pseudo?

src/pivot-table/components/cells/DimensionCell.tsx
```diff
@@ -99,9 +99,7 @@ const DimensionCell = ({
   const { select, isSelected, isActive, isLocked } = useSelectionsContext();
   const isNull = qType === NxDimCellType.NX_DIM_CELL_NULL;
   const selectionCellType = isLeftColumn ? NxSelectionCellType.NX_CELL_LEFT : NxSelectionCellType.NX_CELL_TOP;
-  const isCellLocked =
-    isLocked(selectionCellType, cell.y, colIndex) ||
-    layoutService.isDimensionLocked(selectionCellType, cell.y, colIndex);
+  const isCellLocked = isLocked(selectionCellType, cell.y, colIndex) || cell.isLockedByDimension;
```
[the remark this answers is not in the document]
cbt1: No because `isLocked` is checking if a dimension is locked becaus selections mode is active in another field. So it's not something that Engine is aware of. So it is not available in qDimensionInfo.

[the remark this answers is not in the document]
cbt1: To clarify. A cell should be locked for selections if any of the 2 following conditions are true:
1. Dimension is locked.
2. There are ongoing selections in another field.

src/services/utils/get-sorted-dimension-info.ts
cbt1: In  this logic `qDim.qCardinalities.qHypercubeCardinal > 0` is used as a way to check if a dimension is visible to the user. By visible I mean that a node in it's "parent dimension" must have at least 1 node expanded.

Ex: you have 2 dimensions on the left side. The first dimension does not have any nodes expanded. This means that the second dimension is not visible to the user. But that also means that `qDim.qCardinalities.qHypercubeCardinal > 0` is `false` for the second dimension.

src/services/utils/get-sorted-dimension-info.ts
cbt1: Ideally I want the array to be of type `ExtendedDimensionInfo[]` but in some part of the code, there is a need to iterate the sortedDimensionInfo array and be able to tell if an item in the array is a pseudo dimension or not.

What I think I can do is create a new type that is not a integer but an object for the `PseudoDimension` type instead. That way all items in the array would be objects. But I cannot really make it into a `ExtendedDimensionInfo` type since that has lots of required properties that I don't want to fake for the sake of the solution.

a-m-dev: The thing that may cause problem in that case is even if you have it in object, again when doing a map, filter, ... on the end result, you are going to need to cast iteration item to either that object or `ExtendedDimensionInfo` type and you need to take care of it (but of course it directly ties into how often we want to use that data)

a-m-dev (approved):
LGTM

src/services/utils/get-sorted-dimension-info.ts
cbt1: I'm not seeing any other alternative atm. There is a need to have an array that contains both dimensions and a pseudo dimensions. As both type of dimensions are rendered for the user as a "dimension". I would have to separate that somehow.

I can have an array with only numbers, including `-1`, that points to a qDimensionInfo (essentially the `qEffectiveInterColumnSortOrder`). But when iterating that using `map` or `filter` you end up having to check if the number is `-1` anyway and have some special logic for that. So in that sense the end result is the same or perhaps even more complex.

PR #322, opened 2023-08-11
chore: fix qlik eslint

niekvanstaveren:


src/hooks/use-render.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable react-hooks/rules-of-hooks */
```
cbt1, line 1: Do you know why these could now be removed?

src/pivot-table/components/cells/EmptyCell.tsx
```diff
@@ -25,4 +25,8 @@ const EmptyCell = ({ style, index, isLastRow, isLastColumn, showLastRowBorderBot
   );
 };
 
+EmptyCell.defaultProps = {
```
cbt1, line 28: I think we a better off changing the index to be required.

src/hooks/use-render.ts
enell: That disable isn't needed on main either. I think that is a leftover from some other refactoring

cbt1 (approved):
LGTM, we should update the prettier config and tsconfig after this

niekvanstaveren:
> LGTM, we should update the prettier config and tsconfig after this

Yes, that was my plan as well. Did not want to get this PR too big

PR #328, opened 2023-08-16
feat: head cell menu from shared lib (nebula table utils)

a-m-dev:
This PR imports the `<HeadCellMenu />` from shared lib and uses it for menu. 
~~in order to get this in, [we might have this one merged first](https://github.com/qlik-oss/nebula-table-utils/pull/20) in shared library, until then the ci here will fail because the current package does not includes HeadCellMenu component~~

## what is included:
- [x] `useSorting` hook in supernova: which is responsible for all sorting communication with engine
- [x] storing `activelySortedColumn` on layout of engine which is done by `useSorting`
- [x] Modified `extractHeaders` module to support/include required info on cell/dim/col
- [x] add support for `activelySortedColumn` type on `qHyperCube` of `PivotLayout` interface
- [x] Sorting icon next to dim label
- [x] fix a typing issue that is related to shared lib
- [x] fix header titles ellipses in smaller header width (current behaviour -> rendering test fails)
- [x] add min width of 84px for header and left grid cells (SortIconSize + MinSizeForText + 4px grid gap + MenuIconSize) 
- [x] rename `HeaderTitle` interface to `HeaderCell`

## what is not included: 
- [x] unit tests (next PR to keep this one from growing) (https://github.com/qlik-oss/sn-pivot-table/pull/349)
- [x] clean up 
- [x] refactor some parts if required
- [x] resolve conflicts

###  Please Note:
The big chunk of this PR is related to yarn-lock file because of required packages

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/cb2ccd0c-9830-43fd-bb7e-ecfb92a8d68d


package.json
```diff
@@ -86,6 +86,10 @@
     "@babel/preset-typescript": "7.22.5",
     "@commitlint/cli": "17.7.1",
     "@commitlint/config-conventional": "17.7.0",
+    "@emotion/react": "^11.11.1",
```
cbt1, line 89: Pin the dependencies

src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -1,38 +1,125 @@
...
+    fieldId: cell.fieldId,
+    qLibraryId: cell.qLibraryId,
+    label: "right",
+    headTextAlign: "right" as Align,
+    sortDirection: cell.sortDirection,
+
+    colIdx: cell.colIdx,
+    qReverseSort: cell.qReverseSort,
+    isActivelySorted: cell.isActivelySorted,
+  };
```
a-m-dev, line 76: what/where would be the best to handle this case?

src/hooks/use-sorting.ts
```diff
@@ -0,0 +1,67 @@
...
+interface UseSortingApi {
+  changeSortOrder: ChangeSortOrder;
+  changeActivelySortedHeader: ChangeActivelySortedHeader;
+}
+
+interface UseSorting {
+  (model: Model, hyperCube: HyperCube): UseSortingApi;
+}
+
+const useSorting: UseSorting = (model, qHyperCube) => {
```
cbt1, line 13: This is  not a hook as it's not bound by the rules of hooks.

You should however make it a hook by wrapping the creation of `api` in a `useMemo`.

src/hooks/use-render.ts
```diff
@@ -53,6 +54,7 @@ const useRender = () => {
     [styleService],
   );
   const isFontLoaded = useWaitForFonts(fonts);
+  const { changeSortOrder, changeActivelySortedHeader } = useSorting(model, layout.qHyperCube);
```
cbt1, line 57: Why is this created here? I seems like it can be created in `PivotTable` component instead to avoid a lot of prop drilling?

src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -1,38 +1,129 @@
...
+      </div>
+
+      {isDim && (
+        <>
+          <HeadCellMenu
+            headerData={headerData}
+            translator={translator}
+            tabIndex={-1}
+            anchorRef={anchorRef}
+            handleHeadCellMenuKeyDown={() => console.log("keyDown")}
```
cbt1, line 118: `console.log`?

src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -1,38 +1,129 @@
...
         ...style,
         ...getBorderStyle(true, isLastColumn, styleService.grid.border, false),
         ...styleService.header.rowTitle,
-        display: "flex",
+        padding: 0,
+        position: "relative",
+        display: "grid",
+        gridTemplateColumns: "1fr 24px",
+        gridGap: "4px",
+        minWidth: SORT_ICON_WIDTH + HEADER_TEXT_MIN_WIDTH + MENU_ICON_WIDTH,
```
cbt1, line 98: This fixes the issue where the "hamburger icon" is not aligned vertically.

Before:
Skärmavbild 2023-09-01 kl  08 53 43

After:
Skärmavbild 2023-09-01 kl  08 53 33


```suggestion
        minWidth: SORT_ICON_WIDTH + HEADER_TEXT_MIN_WIDTH + MENU_ICON_WIDTH,
        alignItems: "center",
```

src/hooks/use-render.ts
a-m-dev: I did like this because we are dealing with **supernova work like `useLoadDataPages`** calls to engine. 
I took the same approach as sn-table where putting a barrier between the work that needs to be done in supernova scope and react scope.

src/pivot-table/components/cells/DimensionTitleCell.tsx
a-m-dev: I need to make it optional in shared-lib, for now i'll pass it a `() => {}` function

src/hooks/use-render.ts
cbt1: We actually do call Engine out side of the "supernova scope". Ex: collapsing or expanding a node and fetching more data when the user scrolls.

So I don't really agree with that as a reason.

a-m-dev: moved it to Pivot table component

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -87,7 +87,7 @@ export default function useColumnWidth(
     [leftDimensionData.grid, getLeftColumnWidth],
   );
 
-  const rightGridWidth = useMemo(() => Math.max(rect.width - leftGridWidth - GRID_BORDER), [leftGridWidth, rect.width]);
+  const rightGridWidth = useMemo(() => rect.width - leftGridWidth - GRID_BORDER, [leftGridWidth, rect.width]);
```
cbt1, line 90: Why remove `Math.max`?

src/pivot-table/hooks/use-sorting.ts
```diff
@@ -0,0 +1,71 @@
...
+          return false;
+        }
+      },
+
+      changeActivelySortedHeader: async (header: Header) => {
+        if (!model) throw new Error("No Model provided!");
+
+        const { colIdx, qLibraryId, fieldId, sortDirection } = header;
+        const isActivelySortedColExists = qHyperCube.activelySortedColumn;
+        let patch: EngineAPI.INxPatch[] = [];
```
cbt1, line 46: Looks like `patch` can be declared as a `const`?

src/pivot-table/hooks/use-column-width.ts
a-m-dev: because what it tries to evaluates is just a single value (`rect.width - leftGridWidth - GRID_BORDER`). There is no other value to compare and find the max between them

src/pivot-table/hooks/use-sorting.ts
```diff
@@ -0,0 +1,69 @@
...
+          console.error("Error while applying patch: ", error);
+          return false;
+        }
+      },
+
+      changeActivelySortedHeader: async (header: Header) => {
+        if (!model) throw new Error("No Model provided!");
+
+        const { colIdx, qLibraryId, fieldId, sortDirection } = header;
+        const isActivelySortedColExists = qHyperCube.activelySortedColumn;
```
cbt1, line 45: Guessing the type for `isActivelySortedColExists` is not a boolean?

```suggestion
        const isActivelySortedColExists = !!qHyperCube.activelySortedColumn;
```

cbt1 (approved):
LGTM

PR #329, opened 2023-08-16
build: added eslint rules for react hooks

enell:


src/hooks/use-render.ts
```diff
@@ -36,7 +36,7 @@ const useRender = () => {
   const viewService = useViewService(pageInfo);
   const rect = useSnapshot({ rect: useRect(), layoutService, viewService, model });
   const { qPivotDataPages, isLoading } = useLoadDataPages({ model, layoutService, viewService, pageInfo });
-  const styleService = useMemo(() => createStyleService(theme, layoutService), [theme.name(), layoutService]);
```
cbt1, line 39: It needs to be `theme.name()` because the reference to the `theme` object does not change when a theme is changed.

cbt1 (changes requested):

enell:
> It needs to be `theme.name()` because the reference to the `theme` object does not change when a theme is changed. 

done

src/hooks/use-render.ts
enell: done

cbt1 (approved):
:shipit:

PR #330, opened 2023-08-17
fix: native tooltip on totals cell

cbt1:


src/pivot-table/components/cells/TotalsCell.tsx
```diff
@@ -25,6 +25,7 @@ const TotalsCell = ({ cell, style, isLeftColumn, isLastRow, isLastColumn }: Labe
 
   return (
     <div
+      title={cell.ref.qText}
```
a-m-dev, line 28: I think getting qText from ref can be resulted in undefined right?

cbt1: Yes then the attribute should not be rendered by react

a-m-dev (approved):

PR #331, opened 2023-08-17
feat: use effective properties

cbt1:
Adding a hook to fetch effective properties so that the value `qShowTotalsAbove` can be read. It's needed for the "Total at bottom/top" feature.

### DIFF
--- a/src/hooks/use-effective-properties.ts
+++ b/src/hooks/use-effective-properties.ts
@@ -1,0 +1,14 @@
+import { usePromise } from "@nebula.js/stardust";
+import type { Model, PivotLayout } from "../types/QIX";
+
+const useEffectiveProperties = (model: Model, layout: PivotLayout) =>
+  usePromise<EngineAPI.IGenericObjectProperties | undefined>(async () => {
+    if (layout.snapshotData) {
+      // Is snapshot
+      return {} as EngineAPI.IGenericObjectProperties;
+    }
+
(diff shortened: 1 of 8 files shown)


src/hooks/use-effective-properties.ts
```diff
@@ -0,0 +1,14 @@
+import { usePromise } from "@nebula.js/stardust";
+import type { Model, PivotLayout } from "../types/QIX";
+
+const useEffectiveProperties = (model: Model, layout: PivotLayout) =>
+  usePromise<EngineAPI.IGenericObjectProperties | undefined>(async () => {
+    if (model === undefined || layout.snapshotData || !("getEffectiveProperties" in model)) {
+      // Is snapshot
+      return {} as EngineAPI.IGenericObjectProperties;
```
a-m-dev, line 8: why not return undefined here since we expecting it in types?

a-m-dev (approved):
LGTM, with one minor comment

src/hooks/use-effective-properties.ts
cbt1: @a-m-dev The reason is that `undefined` is returned when the promise have not been resolved yet. So by returning an object, it's possible to determine when the fetch is completed and the chart can be rendered.

Thinking about this some more. The is probably a need to have a `loading` state as well, so that the `layout` and `effective properties` are not out-of-sync when rendering the chart. As the `undefined` state of effective propties only applied the first time the chart is rendered.

PR #332, opened 2023-08-21
feat: useFetch

cbt1:
This PR fixes the issue I mentioned [here](https://github.com/qlik-oss/sn-pivot-table/pull/331#discussion_r1299619397).

Also included:
- A generic `usePromise` mock
- Refactored `useLoadDataPages` to also use the `useFetch` hook.
- `useLoadDataPages` fetched the same data pages an extra time on the first render because to `pageInfo` is updated in a `useEffect`. This PR also fixes that.
- `useLoadDataPages` fetched the same data pages an extra time on first render because when `effectiveProperties` had been fetched, `layoutService` got updated with a new instance. When in reality the hook never used anything other than the `layout` and `isSnapshot` information.

So in short on the "also included". Fetching data pages is now only done 1 time, instead of 3 times on first render.


__mocks__/@nebula.js/stardust.js
```diff
@@ -1,8 +1,14 @@
 const { useState, useMemo, useEffect } = require("react");
 
 const usePromise = (callback) => {
-  callback();
-  return [Promise.resolve(), undefined];
+  const [result, setResult] = useState();
```
cbt1, line 4: I have not looked at the actual `usePromise` code in nebula, but guessing it works something like this. So made sense to have it behave the similar in our tests.

src/hooks/use-render.ts
```diff
@@ -66,7 +66,7 @@ const useRender = () => {
       layoutService &&
       styleService &&
       !!qPivotDataPages &&
-      effectiveProperties &&
```
cbt1, line 69: Removed because `effectiveProperties` can now be `undefined`

src/hooks/use-fetch.ts
```diff
@@ -0,0 +1,19 @@
...
+  const [response, error] = usePromise(async () => {
+    const result = await fetch();
+
+    return {
+      result,
+      deps,
+    };
+  }, deps);
+
+  const isLoading = response === undefined || response.deps.some((dep, idx) => dep !== deps[idx]);
```
a-m-dev, line 16: I think loading state could be incorrect because of: 
1. if the response is void on some rpc/api call we will always have isLoading as true?
2. object coercions on deps array (some crazy deps scenario like below may cause is loading to be true?)

Screenshot 2023-08-22 at 10 15 49


Maybe a solution would be to wrap await part into a try catch and use a state for loading. Then in finally of that try...catch call set that flag to false?

cbt1: 1. Good catch! I think a fix is to simple check if there is an error, than loading is "done". Using state will trigger an extra render when loading is finsished so want to avoid that.
2. I'll look into it

src/hooks/__tests__/use-fetch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      expect(result.current[2]).toBe(undefined);
+    });
+  });
+
+  test("should return is loading state while fetch is completed", async () => {
+    const { result } = renderHook(() => useFetch(() => Promise.resolve(), ["deps"]));
+
+    await waitFor(() => {
+      expect(result.current[0]).toBe(undefined);
+      expect(result.current[1]).toBe(false);
```
a-m-dev, line 20: the above test is exactly like this one, but `result.current[1]` is expected to be true in there. what makes this one to be false?

__mocks__/@nebula.js/stardust.js
```diff
@@ -1,8 +1,32 @@
 const { useState, useMemo, useEffect } = require("react");
 
-const usePromise = (callback) => {
-  callback();
-  return [Promise.resolve(), undefined];
+const usePromise = (callback, deps) => {
```
a-m-dev, line 3: why we are mocking this one? 
it's source exactly doing the same [but in useLayoutEffect](https://github.com/qlik-oss/nebula.js/blob/91e04c9a0e1df94cf17edeed6dd9603e7b6de03e/apis/supernova/src/hooks.js#L411-L456)

a-m-dev (approved):

__mocks__/@nebula.js/stardust.js
cbt1: The original version can only be run in supernova context. So we need to mock it.

src/hooks/__tests__/use-fetch.test.ts
cbt1: Only time actually. So the first runder it will be `true` and the second render it will be `false`

PR #334, opened 2023-08-23
feat: total divider

cbt1:
Adds a divider for the first total row and column.

Before:
Skärmavbild 2023-08-23 kl  07 28 26

After:
Skärmavbild 2023-08-23 kl  07 28 04


src/pivot-table/hooks/use-is-total-cell.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    [topDimensionData],
+  );
+
+export const useIsTotalValue = (leftDimensionData: LeftDimensionData, topDimensionData: TopDimensionData) =>
+  useCallback(
+    (x: number, y: number) =>
+      !!(
+        topDimensionData.grid[topDimensionData.grid.length - 1]?.[x]?.isTotal ||
+        leftDimensionData.grid[leftDimensionData.grid.length - 1]?.[y]?.isTotal
```
a-m-dev, line 50:
```suggestion
        topDimensionData.grid.at(-1)?.[x]?.isTotal ||
        leftDimensionData.grid.at(-1)?.[y]?.isTotal
```

it should be same for `lastColumnIndex` in two above functions as well

src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -1,6 +1,11 @@
 import NxDimCellType from "../../../types/QIX";
 import type { Cell, VisibleDimensionInfo } from "../../../types/types";
 
+// qElemNo === -1 => Total
+// qElemNo === -2 => Null
+// qElemNo === -3 => Others
+// qElemNo === -4 => Empty
+
```
a-m-dev, line 8: extra comment maybe?

src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -10,26 +15,49 @@ const createCell = (
...
+    // A getter because child nodes are added as cells are being created. It has to be resolved when it's called.
+    get isLastChild(): boolean {
+      // Root is considedered last child for the total cell divider use case
+      if (root === null) {
+        return true;
+      }
+
+      // Having "parent.isLastChild" means that all ancestors also have to be the last child,
+      // which is needed for the Total cell divider use case.
+      return parent?.children[parent.children.length - 1] === cell && parent.isLastChild;
```
a-m-dev, line 54: is `cell` refers to the current cell object that is being created? then how this condition could be possible? 
```tsx
parent?.children[parent.children.length - 1] === cell
```

even if both objects will having the same key/values, it will be false because cell is being generated every time that js interpreter reaches this line. (Or am i missing something?)

Screenshot 2023-08-29 at 10 32 07

src/pivot-table/components/cells/EmptyCell.tsx
```diff
@@ -1,24 +1,47 @@
 import React from "react";
+import type { Cell } from "../../../types/types";
 import { useStyleContext } from "../../contexts/StyleProvider";
-import { getBorderStyle } from "../shared-styles";
+import { shouldShowTotalCellDivider } from "../../hooks/use-is-total-cell";
+import { getBorderStyle, getTotalCellDividerStyle } from "../shared-styles";
 
 export interface EmptyCellProps {
+  cell: Cell | undefined;
```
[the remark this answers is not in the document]
cbt1: Yes, `EmptyCell` will get an lint error for `react/require-default-props` but I'll just disable that for now

src/pivot-table/data/helpers/create-cell.ts
cbt1: I think the commet for `isLastChild` tries to describe why: "// A getter because child nodes are added as cells are being created. It has to be resolved when it's called."

I simplified version of it should look something like this:
```js
const parent = { isLastChild: true, children: [] );
const firstCell = { isLastChild: () => parent.children.at(-1) === firstCell && parent.isLastChild };
const secondCell = { isLastChild: () => parent.children.at(-1) === secondCell && parent.isLastChild };

parent.children.push(firstCell);
firstCell.isLastChild(); // => TRUE, but this is not when it's executed.

parent.children.push(secondCell);
// All cells have been added. The cells are now rendered and "isLastChild" is called.
firstCell.isLastChild(); // => FALSE
secondCell.isLastChild(); // => TRUE
```

Does that clear it up a little bit?

cbt1: Or maybe you are thinking about closure and scope?

cbt1: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures#closure

a-m-dev: Yes, it is clean now. 

Basically what i understood is: the cell within each of `isLastChild` function (which exists within the same cell) creates a closure that holds a reference to the same cell. 

and when you execute that function, it checks the same reference that it's closure holds.

What i was confused was with the checking part with cell (`=== cell` part), then I questioned my self why not using `this` here which is more not confusing (at least for me). 

But now i figured out why since you are using arrow function, `this` will point to window object (I believe) and not the correct reference to `cell` object. But it works if you convert `isLastChild` to typical `function() {...}` call

but i'll leave that judgment up to you to see which one is better to go with in this case, and of course its LGTM now 🚢

a-m-dev (approved):

PR #344, opened 2023-08-28
chore: remove mashup example

cbt1:
Mashup examples are available in the general [nebula mashup repo](https://github.com/qlik-oss/nebula.js-examples/tree/main) and in the [qlik.dev](https://qlik.dev/embed/visualizations-to-embed/pivot-table/sn-pivot-table) portal


a-m-dev:
is there any specific reason why we want to remove examples? because AFAIK sn-table [still keeps mashup example](https://github.com/qlik-oss/sn-table/tree/main/mashup-example) in the repo as well.

cbt1:
> is there any specific reason why we want to remove examples? because AFAIK sn-table [still keeps mashup example](https://github.com/qlik-oss/sn-table/tree/main/mashup-example) in the repo as well.

We've said that they should be removed from sn-table as well. So that all examples are at a single source (+ qlik.dev)

a-m-dev (approved):

PR #346, opened 2023-08-28
chore: update nebula-table-utils to 2.0.0

enell:


package.json
```diff
@@ -92,9 +92,9 @@
     "@nebula.js/cli-sense": "4.4.0",
     "@nebula.js/cli-serve": "4.4.0",
     "@playwright/test": "1.37.1",
-    "@qlik-oss/nebula-table-utils": "1.1.0",
     "@qlik-trial/sprout": "3.4.0",
     "@qlik/eslint-config": "0.4.15",
+    "@qlik/nebula-table-utils": "^2.0.0",
```
cbt1, line 97: Pin the version

a-m-dev: Yes, pls, then it should be LGTM 🚀
Also sn-table needs the same PR as well otherwise it's CI may fail in installation step i guess (ot it still will continue to use the existing old pkg from github registry)

enell: Done

It is true that sn-table needs the same update but I havn't removed anything so nothing will break there

yarn.lock
```diff
@@ -2008,6 +2003,11 @@
     eslint-plugin-vitest "0.2.8"
     svelte-eslint-parser "0.32.2"
 
+"@qlik/nebula-table-utils@^2.0.0":
```
cbt1, line 2006: You forgot to run `yarn` when you pinned the version

a-m-dev (approved):

PR #351, opened 2023-09-04
fix: wrong data after scroll

cbt1:
Primarly fixes an issue that could happen when you scrolled in a pivot table with expanded nodes. Cells could sometimes be rendered at the wrong position.

But also fixes an issue with `distanceToNextCell` in combination with paging.

Notice how the column with pseudo dimensions is not render correctly or the total value row in the last column
Skärmavbild 2023-09-05 kl  15 05 00

Changes includes:
- Re-write of how `leafCount` is calculated. The value is now page based.
- `distanceToNextCell` was mixing page based coordinates and dataset based coordinates. This is now "fixed" so that it always uses page based coordinates.
- When fetching a cell that already exist in the data. The cell is not re-created. This fixes an issue where `cell.children` would only contain the children from the last data that was fetched.
- The fix of `leafCount` broke the total cell divider logic. So that is now refactored into some hopefully less complex.


src/types/types.ts
```diff
@@ -75,18 +76,18 @@ export interface Cell {
   y: number; // y position of cell in dataset
   pageX: number; // X position of cell in page
   pageY: number; // Y position of cell in page
+  mainAxisPageCoord: number; // Either equal pageX or pageY depending on if a cell is in the left or top grid
```
cbt1, line 79: Borrowed the "main axis" part from css flexbox. Maybe bit too contrived name, but very useful.

src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -1,11 +1,46 @@
...
 import type { Cell, VisibleDimensionInfo } from "../../../types/types";
+import { MAX_COLUMN_COUNT, MAX_ROW_COUNT } from "../../constants";
 
 // qElemNo === -1 => Total
 // qElemNo === -2 => Null
 // qElemNo === -3 => Others
 // qElemNo === -4 => Empty
 
+const countLeafNodes = (count: number, node: EngineAPI.INxPivotDimensionCell): number =>
+  node.qSubNodes.reduce((acc, childNode) => {
```
haxxmaxx, line 11: what does acc stand for? accumulation?

src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -24,38 +60,34 @@ const createCell = (
...
-      // which is needed for the Total cell divider use case.
-      return parent?.children.at(-1) === cell && parent.isLastChild;
-    },
   };
 
-  parent?.children.push(cell);
+  if (parent) {
+    const pageDir = isLeftColumn ? "pageY" : "pageX";
+    // eslint-disable-next-line no-param-reassign
+    parent.children[cell[pageDir] - parent[pageDir]] = cell;
```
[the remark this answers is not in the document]
cbt1: Yes, I just did not want to touch this PR if someone was actively reviewing it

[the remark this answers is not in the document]
haxxmaxx: fine by me to change it now :)

src/pivot-table/data/helpers/assign-distance-to-next-cell.ts
```diff
@@ -1,28 +1,20 @@
...
-          // This is what enables the dimensions with branch nodes to be fully scrollable.
-          // By "faking" the distanceToNextCell for the last cell to include all other cells
-          // the react-window list can render with a full size.
-          const sizeDirectionKey: keyof Point = dirKey === "pageY" ? "y" : "x";
-          cell.distanceToNextCell = size[sizeDirectionKey] - (cell[dirKey] + cell.leafCount);
-        }
+        // This is what enables the dimensions with branch nodes to be fully scrollable.
+        // By "faking" the distanceToNextCell for the last cell to include all other cells
+        // the react-window list can render with a full size.
+        cell.distanceToNextCell = size - (cell.mainAxisPageCoord + cell.leafCount);
```
a-m-dev, line 17: is `cell.leafCount` implies only the leafs that are visible in current page?

src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -1,11 +1,46 @@
...
+  // If a node is at the end of a page and has more child nodes at the next page.
+  // Those child nodes should not me included in the leaf count.
+  const maxLeafCount = isLeftColumn ? MAX_ROW_COUNT - pageY : MAX_COLUMN_COUNT - pageX;
+  let nbrOfLeafNodesThatCanBeFetched = node.qUp + node.qDown;
+  // Leaf count is per PAGE, so if the node is the first node on a page. Ignore any nodes on previous page.
+  if (isLeftColumn && pageY === 0) {
+    const fetchTop = y + node.qUp; // qTop value when node was fetched
+    const nbrOfRowsToFirstRowOnPage = fetchTop % MAX_ROW_COUNT;
+    nbrOfLeafNodesThatCanBeFetched = node.qDown + nbrOfRowsToFirstRowOnPage;
+  }
```
a-m-dev, line 39: shouldn't we just say: 
```
nbrOfLeafNodesThatCanBeFetched = node.qDown
```

Because we start from the beganing of the page? because in that case what matters to us is everything below the node right (`qDown`, probably with a - 1 that subtracts the current node)

a-m-dev (commented):
Generally LGTM, just two questions i asked

src/pivot-table/data/helpers/assign-distance-to-next-cell.ts
cbt1: Yes. It could also be named `leafCountOnPage`

src/pivot-table/data/helpers/create-cell.ts
cbt1: We are not always starting from the "top" of the page. If a user scrolls down to some y position and expand or collapse a node there. The starting postion is going to be in proximity to where that node was expanded/collapsed.

Which is why `nbrOfRowsToFirstRowOnPage` is needed.

a-m-dev: but within this if check you have `pageY === 0`, so basically for every page first cell this code will execute and you still calculate `nbrOfRowsToFirstRowOnPage` based on `fetchTop` that uses `y`

or am i missing something maybe?

a-m-dev (approved):

PR #354, opened 2023-09-05
refactor: smarter args for measure text

haxxmaxx:
Basically breaking out this: https://github.com/qlik-oss/sn-pivot-table/pull/287/files#diff-ec18bf142eff172651b535422a1a8fa75c79d87588adbc461acef39767dc630eR42-R43. Will be more needed as of that PR, since we take more styling into account.

Passing the whole styling object since we always use the size and fontFamliy from a specific part of the chart (rowContent, header etc). This would also simplify if we would need other styling as well, potentially bold for example

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -222,6 +222,8 @@ interface DimensionContentStyling extends FontStyling {
   measureLabel: CellStyling;
 }
 
+export type MeasureTextStyling = HeaderStyling | MeasureContentStyling | DimensionContentStyling;
+
 interface GridStyling {
   rowHeight: "compact";
   lineCount: number;
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -222,8 +222,6 @@ interface DimensionContentStyling extends FontStyling {
   measureLabel: CellStyling;
 }
 
-export type MeasureTextStyling = HeaderStyling | MeasureContentStyling | DimensionContentStyling;
-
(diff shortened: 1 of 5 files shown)


src/types/types.ts
```diff
@@ -222,6 +222,8 @@ interface DimensionContentStyling extends FontStyling {
   measureLabel: CellStyling;
 }
 
+export type MeasureTextStyling = HeaderStyling | MeasureContentStyling | DimensionContentStyling;
```
cbt1, line 225: Can the type not be `{ fontSize: string; fontFamily: string }`? That would make it easier when we move this hook to the shared util repo.

Or we might end up moving the measure text hook from sn-table into the util repo instead. But still, the goal should be that both repos should use the same measure hook.

cbt1 (approved):
LGTM just a minor comment about sharing the hook

cbt1:
FYI those rendering tests are failing on main as well. So I think we can safely merge this PR anyway

src/types/types.ts
haxxmaxx: good point, I will have a look at sn-table so the typing and functionality match

a-m-dev (approved):

a-m-dev:
I also had the same issue in my test pr with only 2 (new) rendering tests screenshot. 
So I think you need update your pr with those screenshots and merge it.

PR #356, opened 2023-09-08
feat: base provider

cbt1:


haxxmaxx:
Feels like these props are passed to components which shouldn't be needed anymore? Are you starting to use the provider in the next PR?

haxxmaxx:
sorry wrong button

cbt1:
> Feels like these props are passed to components which shouldn't be needed anymore? Are you starting to use the provider in the next PR?

Yes. Two new PR:s are coming after this one.

First which enabled the header menu selections. It's using the `app`, `model` and `interactions` on a component cell level. So lots of prop drilling there.

Second is a refactoring where `constraints` is replace with `interactions`. Which affects the `DimensionCell` component an all it's parent components.

a-m-dev (approved):

PR #357, opened 2023-09-08
refactor: simpler logic for border width

haxxmaxx:
Since there will be `showLastColumnBorderRight` arg to the border func added with the column width, it started to become a lot of if statements. Turns out you could simplify it quite a lot


cbt1 (approved):
❤️‍🔥

PR #358, opened 2023-09-08
feat: header menu selections

cbt1:
Skärmavbild 2023-09-08 kl  12 32 07

### DIFF
--- a/src/pivot-table/components/cells/DimensionTitleCell.tsx
+++ b/src/pivot-table/components/cells/DimensionTitleCell.tsx
@@ -126,6 +126,7 @@ const DimensionTitleCell = ({
             translator={translator}
             tabIndex={-1}
             anchorRef={anchorRef}
+            handleHeadCellMenuKeyDown={() => {}}
             menuAvailabilityFlags={FLAGS}
             sortRelatedArgs={sortRelatedArgs}
             searchRelatedArgs={searchRelatedArgs}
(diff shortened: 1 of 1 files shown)


src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -75,10 +85,18 @@ const DimensionTitleCell = ({
     [cell, isDim],
   );
 
-  const sortFromMenu = async (evt: React.MouseEvent, newSortDirection: SortDirection) => {
-    evt.stopPropagation();
-    await changeSortOrder(headerData, newSortDirection);
-  };
+  const sortRelatedArgs = useMemo(() => {
```
cbt1, line 88: Might be a bit overkill to wrap these props in `useMemo`. I wrapped the `HeadCellMenu` component in the util repo with `React.memo`. Think I need some more testing to ensure that it actually reduces the number of commited renders.

a-m-dev: +1

src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -45,6 +47,13 @@ const labelTextStyle: React.CSSProperties = {
 
 export const testId = "title-cell";
 
+const handleHeadCellMenuKeyDown = () => {};
```
a-m-dev, line 50: This is optional now with use of (v2.0.1). you can skip it

a-m-dev (approved):
🚀

src/pivot-table/components/cells/DimensionTitleCell.tsx
cbt1: It did not change the number of renders. I've now removed useMemo.

PR #366, opened 2023-09-11
feat: `<HeadCell />` as button

a-m-dev:
https://github.com/qlik-oss/sn-pivot-table/assets/29652890/a22b14dd-4f8b-4e3d-9a31-f7f8ee7837b8

it also works with custom theme as well, it's just the opacity is so low that you barely notice it

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/eb1a8a1a-e6a8-44c4-9fa2-bb5b6ca27bd6


cbt1:
The hovering color will not work when the background color is dark. Worse case scenario being black. What we should do is identify if the color is a dark color, if so, make it brigther, if not, make it darker. By changing the alpha value.

But for this PR the current solution is good enough. The state logic is what we want.

cbt1:
Can you change the "default" theme in `nebula.config.js` so that both `header.rowTitle.background` and `header.columnTitle.background` is white?

src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -110,9 +124,14 @@ const DimensionTitleCell = ({
         gridTemplateColumns: "1fr 24px",
         gridGap: "4px",
         alignItems: "center",
+        cursor: "pointer",
```
cbt1, line 127: This should be a conditional cursor and only be a pointer if `interactions.active` is `true`

src/services/style-service.ts
```diff
@@ -86,7 +86,7 @@ const createStyleService = (theme: ExtendedTheme, layoutService: LayoutService):
         background:
           resolveColor(theme, headerStyling?.[Path.RowTitle]?.[Attribute.Background]) ??
           getThemeStyle([Path.Header, Path.RowTitle], Attribute.Background) ??
-          Colors.Black3,
+          Colors.White,
```
cbt1, line 89: This should be `transparent`

src/pivot-table/hooks/use-head-cell-dim.ts
```diff
@@ -0,0 +1,22 @@
...
+export const useHeadCellDim = ({ open }: UseHeadCellDim) => {
+  const [isHovered, setIsHovered] = useState(false);
+
+  useEffect(() => {
+    if (!open) setIsHovered(false);
+  }, [open]);
+
+  const shadeOpacity = useMemo(() => {
+    let opacity = 0;
+    if (isHovered) opacity = 0.03;
```
haxxmaxx, line 16: why not do this with css?

cbt1: See my comment about "The hovering color...", I don't there is way to achieve that with CSS. If there is that would be great tho.

a-m-dev: I think i've accomplished what we want here with [sass color functions](https://sass-lang.com/documentation/modules/color/) + [sass flow control rules](https://sass-lang.com/documentation/at-rules/control/if/) before, where it just automatically could detect for example text color based on background color. 

But the only reason I choose to do it like so is because in PVT, we are using inline styling which makes it hard for us to detect the element states like `:hover`, `:active` and ... in pure jss style

src/pivot-table/components/cells/DimensionTitleCell.tsx
a-m-dev: is this what we want right? 
(I'll fix menu icon in shared lib asap)

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/84541520-dd33-4f09-9ba5-b0ed3e7a5465

cbt1: Yes 👍

a-m-dev: Promised PR is here to review: https://github.com/qlik-oss/nebula-table-utils/pull/46

cbt1 (approved):

src/pivot-table/hooks/use-head-cell-dim.ts
haxxmaxx: I see. You can do it using styled-components. Have we decided anything there what we should do in this repo? Since a lot of things use styled components in sn-table, but not in pivot table

cbt1: The main concern with styled components is performance when scrolling. Since this particular component is rather static in nature, it should be "safe" to use styled components here. Which can solve the "hover" part of the feature

PR #367, opened 2023-09-12
chore: remove sort icon

a-m-dev:


src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -132,11 +130,6 @@ const DimensionTitleCell = ({
     >
       <div style={{ ...headCellBackgroundDim, opacity: shadeOpacity }} />
       <div style={{ ...labelWrapperStyle }}>
```
a-m-dev, line 132: we can remove this wrapper as well, but let's keep it for lock icon

cbt1 (approved):

PR #370, opened 2023-09-13
feat: header hover background

cbt1:
Requires this https://github.com/qlik-oss/nebula-table-utils/pull/49/ to work.

Gets the hover background color from the util repo and uses styled component to trigger the hover effect.


src/pivot-table/hooks/use-head-cell-dim.ts
```diff
@@ -1,22 +1,14 @@
...
+import { useState } from "react";
 
 interface UseHeadCellDim {
-  open: boolean;
+  interactions: stardust.Interactions;
 }
 
-export const useHeadCellDim = ({ open }: UseHeadCellDim) => {
-  const [isHovered, setIsHovered] = useState(false);
+export const useHeadCellDim = ({ interactions }: UseHeadCellDim) => {
```
a-m-dev, line 8: Does your changes takes care of active state? 
hover state should be 3% and active state is 5% darker

cbt1: Currently no, it's roughly 7% for both hover and active. I will update the util repo so that its ~3% and ~5%. But while testing that on my screen I cannot see that color difference with only 3%, unless it goes from white to darker. 

Something like a purple color will most likely need UX calibrated screens to see any difference. But I'll stick with the values they provided anyway.

Good find!

src/pivot-table/components/cells/styles.tsx
```diff
@@ -0,0 +1,31 @@
...
+import type { stardust } from "@nebula.js/stardust";
+
+interface Props {
+  interactions: stardust.Interactions;
+  background: string;
+  hoverBackground: string;
+}
+
+export const StyledHeaderCellWrapper = styled(Box, {
+  shouldForwardProp: (prop: string) => prop !== "interactions" && prop !== "hoverBackground" && prop !== "background",
```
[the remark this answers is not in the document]
cbt1: No I don't think so

test/rendering/render.mjs-snapshots/scenario-1-chromium-linux.png
cbt1, line 1: Not entire sure why the labels are updated but guessing that the new HeadCellMenu got something to do with that

a-m-dev (approved):
🚀

PR #371, opened 2023-09-13
fix: columns in left grid out of sync

cbt1:
Uses the actual number of rows in a page instead of the maximum possible when computing some values. I ended up putting the value for that on `PageInfo` as a way to make it easily shared across components/hooks.

Issue that have been fixed:
Skärmavbild 2023-09-13 kl  11 31 02

### DIFF
--- a/src/pivot-table/components/grids/LeftGrid.tsx
+++ b/src/pivot-table/components/grids/LeftGrid.tsx
@@ -1,7 +1,7 @@
 import { useOnPropsChange } from "@qlik/nebula-table-utils/lib/hooks";
 import React, { memo, useLayoutEffect } from "react";
 import { VariableSizeList } from "react-window";
-import type { DataModel, LayoutService, LeftDimensionData, VisibleDimensionInfo } from "../../../types/types";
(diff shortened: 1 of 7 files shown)


src/pivot-table/components/grids/LeftGrid.tsx
```diff
@@ -75,7 +77,7 @@ const LeftGrid = ({
     }
   }, [getScrollTop, layoutService, leftGridRef]);
 
-  const totalHeight = layoutService.size.y * contentCellHeight;
+  const totalHeight = pageInfo.rowsOnCurrentPage * contentCellHeight;
```
cbt1, line 80: This was a root cause

src/pivot-table/components/grids/LeftGrid.tsx
```diff
@@ -89,7 +91,7 @@ const LeftGrid = ({
         const { itemCount, estimatedItemSize, listValues } = getListMeta(
           list,
           totalHeight,
-          layoutService.size.y,
+          pageInfo.rowsOnCurrentPage,
```
cbt1, line 94: This was a root cause

PR #375, opened 2023-09-19
feat: pagination footer

cbt1:


src/pivot-table/components/Wrapper.tsx
```diff
@@ -1,30 +1,47 @@
...
+        pageInfo={pageInfo}
+        totalRowCount={pageInfo.totalRowCount}
+        totalColumnCount={0}
+        totalPages={pageInfo.totalPages}
+        keyboard={keyboard}
+        translator={translator}
+        theme={theme}
+        interactions={interactions}
+        rect={rect as stardust.Rect}
+        layout={layout as unknown as EngineAPI.IGenericHyperCubeLayout}
```
a-m-dev, line 43: I think you don't need to convert it to unknown first since it's happy with only `as EngineAPI.IGenericHyperCubeLayout`. 

I got no type complain on it

src/pivot-table/hooks/__tests__/use-table-rect.test.ts
```diff
@@ -34,13 +35,13 @@ describe("use-table-rect", () => {
   test("should consider pagination height", () => {
     shouldShowPagination = true;
     const { height } = renderer();
-    expect(height).toBe(rect.height - PAGINATION_HEIGHT);
+    expect(height).toBe(rect.height - PAGINATION_HEIGHT - 1);
```
a-m-dev, line 38: why is that extra 1 px?
is it a border somewhere in pagination? is so, i think it's better to use the exported value of `PAGINATION_FOOTER_BORDER` instead

src/types/types.ts
```diff
@@ -141,14 +141,6 @@ export interface ExtendedSelections extends stardust.ObjectSelections {
   removeListener: (name: string, callback: () => void) => void;
 }
 
-export interface ExtendedTheme extends stardust.Theme {
-  name: () => string;
```
a-m-dev, line 145: btw, I've noticed that we still keep `name()` property [at here](https://github.com/qlik-oss/nebula-table-utils/blob/6003a8a04a64655410b95131ab04714dc0f175b8/src/hooks/use-extended-theme/types.ts#L11), but stardust Theme has it already

just as a reminder for a next small refactor pr in shared lib

Screenshot 2023-09-21 at 12 29 46

a-m-dev (approved):
Looks great 🚀
just left some minor comments

src/hooks/use-render.ts
```diff
@@ -31,11 +31,13 @@ const useRender = () => {
...
   const app = useApp();
   const embed = useEmbed();
+  const keyboard = useKeyboard();
   const interactions = useInteractionState();
   const [effectiveProperties, isLoadingEffectiveProperties] = useEffectiveProperties(model, layout);
   const layoutService = useLayoutService(layout, effectiveProperties);
   const selections = useSelections() as ExtendedSelections;
-  const theme = useTheme() as ExtendedTheme;
+  const theme = useExtendedTheme(element);
+  const themeName = theme.name();
```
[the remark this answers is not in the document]
cbt1: Yes, wanted to keep the lint rule. Otherwise the linter will complain that `theme` is missing in the dependency, because you have to use `theme.name()` to catch changes

src/pivot-table/components/Wrapper.tsx
cbt1: Will take a look at it

src/pivot-table/hooks/__tests__/use-table-rect.test.ts
cbt1: Yes that 1 is the border. Will fix and use const value instead

PR #379, opened 2023-09-21
chore: locked cells now keeps the original background color

a-m-dev:
https://github.com/qlik-oss/sn-pivot-table/assets/29652890/7ecd4fd0-d5da-47a9-9ba9-413fb584b146


cbt1 (approved):
❤️‍🔥

PR #381, opened 2023-09-25
fix: load out of range data page

cbt1:
Fixes an issue where the chart could crash when a node was collapsed the total number of columns/rows was greatly reduced.

Also fixes so that `useLoadPagePages` does not try to fetch data for the "next" page.

Before:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/4d6531ad-64c7-4264-9421-40574674ee78

After:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/b71c2ebc-c8fc-4c93-9cd1-273d8276da6b

### DIFF
--- a/src/hooks/use-load-data-pages.ts
+++ b/src/hooks/use-load-data-pages.ts
@@ -59,7 +59,7 @@ export const getFetchArea = (
   }
 
   let qLeft = 0;
-  let qTop = 0;
+  let qTop = pageStartTop;
 
   // qLastExpandedPos only exist in the layout if a new layout was received because a node was expanded or collapsed
   if (qLastExpandedPos) {
--- a/src/hooks/use-load-data-pages.ts
+++ b/src/hooks/use-load-data-pages.ts
@@ -1,5 +1,6 @@
 import { useFetch } from "@qlik/nebula-table-utils/lib/hooks";
 import { DEFAULT_PAGE_SIZE, Q_PATH } from "../constants";
+import { MAX_COLUMN_COUNT } from "../pivot-table/constants";
 import type { Model, PivotLayout } from "../types/QIX";
 import type { LayoutService, PageInfo, ViewService } from "../types/types";
 
@@ -10,18 +11,12 @@ interface Props {
   pageInfo: PageInfo;
 }
 
-export const shouldFetchAdditionalData = (
-  qLastExpandedPos: EngineAPI.INxCellPosition | undefined,
-  viewService: ViewService,
-) => {
+export const shouldFetchAdditionalData = (qLastExpandedPos: EngineAPI.INxCellPosition | undefined) => {
   if (!qLastExpandedPos) {
     return false;
   }
 
-  return (
-    viewService.gridColumnStartIndex + viewService.gridWidth >= DEFAULT_PAGE_SIZE ||
-    viewService.gridRowStartIndex + viewService.gridHeight >= DEFAULT_PAGE_SIZE
-  );
+  return qLastExpandedPos.qx >= DEFAULT_PAGE_SIZE || qLastExpandedPos.qy >= DEFAULT_PAGE_SIZE;
 };
 
 export const isMissingLayoutData = (layout: PivotLayout, pageInfo: PageInfo): boolean => {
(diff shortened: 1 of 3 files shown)


src/hooks/use-load-data-pages.ts
```diff
@@ -36,6 +37,46 @@ export const isMissingLayoutData = (layout: PivotLayout, pageInfo: PageInfo): bo
...
+    return null;
+  }
+
+  let qLeft = 0;
+  let qTop = pageStartTop;
+
+  // qLastExpandedPos only exist in the layout if a new layout was received because a node was expanded or collapsed
+  if (qLastExpandedPos) {
+    // gridColumnStartIndex might not exist anymore in the new expanded/collapsed layout
+    qLeft = Math.max(0, Math.min(qSize.qcx - DEFAULT_PAGE_SIZE, viewService.gridColumnStartIndex));
```
a-m-dev, line 67: why we are subtracting `DEFAULT_PAGE_SIZE` from `qSize.qcx`? 
let's look at the data from end of it in x axis:
let's say we have 1000 cols
and now we are at the last 50 col range (something like 980 as `gridColumnStartIndex`)

the `qLeft` will always endup being `qSize.qcx - 50` and always will try to fetch from index 950 when user scrolls within 950 to 1000 col. 

now if user once triggered fetched from (let's say) index 930 -> and now scrolls to the end of it's avaliable dataset (980) -> it would definitly require one more fetch to get the last 20 cols -> but this time -> it will try to fetch last 50 because of `Math.min(qSize.qcx - DEFAULT_PAGE_SIZE, viewService.gridColumnStartIndex)` 

Maybe i'm overthinking it but do we need to care about that kind of scenario? 
(Probably we have it already when data overlaps in between scrolls -> at least in y axis)

src/hooks/use-load-data-pages.ts
```diff
@@ -36,6 +37,46 @@ export const isMissingLayoutData = (layout: PivotLayout, pageInfo: PageInfo): bo
...
+ * guaranteed to work with a new layout, as its properties may no longer be valid.
+ */
+export const getFetchArea = (
+  qLastExpandedPos: EngineAPI.INxCellPosition | undefined,
+  viewService: ViewService,
+  qSize: EngineAPI.ISize,
+  pageInfo: PageInfo,
+) => {
+  const pageStartTop = pageInfo.page * pageInfo.rowsPerPage;
+  const pageEndTop = pageStartTop + pageInfo.rowsPerPage;
```
a-m-dev, line 51: if we sum start of the page with rows per page, then it should not be pageEndTop -> probably something like pageEnd(index?) should be good maybe (or any other suggestion)

also pageStartIndex instead of pageStartTop maybe?

src/hooks/use-pagination.ts
```diff
@@ -17,7 +17,7 @@ const getRowsOnCurrentPage = ({ rowsPerPage, totalRowCount, page }: PageInfo) =>
   Math.min(rowsPerPage, totalRowCount - page * rowsPerPage);
 
 const getPageMeta = (qcy: number, page: number) => {
-  const rowsPerPage = Math.min(qcy, MAX_ROW_COUNT);
+  const rowsPerPage = MAX_ROW_COUNT;
```
a-m-dev, line 20: what if we have values less than MAX_ROW_COUNT?, which is more probable with smaller datasets. 
will it affect the calculations in `use-load-data-pages.ts` file?

a-m-dev (commented):
Since `rowsPerPage` is now `MAX_ROW_COUNT` and it plays a big role in `use-load-data-pages.ts` file to calculate `pageStartTop` and `pageStartEnd` variables, I think this changes may effect any dataset that has less rows than `MAX_ROW_COUNT`. 

have you tried to test it in that scenario?

src/hooks/use-load-data-pages.ts
cbt1: This hook does not get called when the user scrolls. It will fetch data if the layout changes, either because of some prop change, or as in the code line, the user have expanded/collapsed a node.

Loading data on scroll is handled else where and may or may not fetch overlapping data. But that logic has not changed.

src/hooks/use-pagination.ts
cbt1: Yes, I changed this intentionally. `PageInfo` is out-of-sync with the new layout (on the first render). So If we use the value from `PageInfo.rowsPerPage` it would be the size of the previous layout.

The easiest fix was to just set this as `MAX_ROW_COUNT` since `rowsOnCurrentPage` does hold that info anyway and works better on all other pages.

cbt1: So to clarify. Previously when the user scrolled to the end and expanded a node at row 95 out of 100 rows. A new layout is recieved. `PageInfo.rowsPerPage` will be `100` but `layout...qSize.qcy` might be something like `130`. This would mean that a fetch is sent for to only fetch data up until row 100. But it now actualy have 30 more rows that should be fetched. 

Then on the next render when the `PageInfo` have had it's state updated via a `useEffect`, another fetch is triggerd. This time with the correct value on `rowsPerPage`.

cbt1: So this code change actually make so that data is only fetched once and with the correct number of rows

a-m-dev (approved):

PR #383, opened 2023-09-26
fix: scroll to bottom not loading cell

cbt1:
Fixes an issue where the last cell in the first column would not always be visible when scrolling to the bottom after expanding a cell.

Also fixes an issue where "grid border" was not included in some places, causing the `scrollTo` call for the react-window components to never reach the end of their scroll (bottom or right). It's was just 1 px of so there was likely no visible effect.

Before (notice how 2007 is not visible aftar expand "Mar"):

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/8cd188ac-d493-451d-94cd-cc40ff84b184

After:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/a94083b4-4260-4e53-87c5-f0c565b24d0e


src/pivot-table/components/helpers/get-item-size-handler.ts
```diff
@@ -16,7 +16,7 @@ export const getRowHeightHandler =
     const cell = isLastColumn ? list[rowIndex] : Object.values(list)[rowIndex];
 
     if (rowIndex === 0 && cell?.pageY > 0) {
-      return (cell.leafCount + cell.pageY) * cellHeight;
+      return (cell.leafCount + cell.pageY + cell.distanceToNextCell) * cellHeight;
```
cbt1, line 19: This is the actual fix

test/rendering/render.mjs-snapshots/scenario-1-chromium-linux.png
cbt1, line 1: Snapshots updated because a 1 px bug was fixed. See below "red dot".

Skärmavbild 2023-09-27 kl  07 19 24

a-m-dev (approved):

PR #387, opened 2023-09-28
fix: change page while loading data

cbt1:
Fixes an issue where data from previous page could end up on the new page.

It could be reproduces in the following way:
1. Go to last page
2. Scroll down to bottom. Should be more than 50 rows on the page and the call to `getHyperCubePivotData` have to be slow.
3. Go to first page.

Now the data from the last page would be rendered on the first page.

### DIFF
--- a/src/pivot-table/hooks/use-data-model.ts
+++ b/src/pivot-table/hooks/use-data-model.ts
@@ -3,6 +3,7 @@ import { useCallback, useMemo } from "react";
 import { Q_PATH } from "../../constants";
 import type { Model } from "../../types/QIX";
 import type { DataModel, ExpandOrCollapser, FetchMoreData, PageInfo } from "../../types/types";
+import useMutableProp from "./use-mutable-prop";
 
 export interface UseDataModelProps {
   model: Model;
@@ -11,6 +12,7 @@ export interface UseDataModelProps {
 }
 
 export default function useDataModel({ model, nextPageHandler, pageInfo }: UseDataModelProps): DataModel {
+  const currentPage = useMutableProp(pageInfo.page);
   const genericObjectModel = model as EngineAPI.IGenericObject | undefined;
 
   const collapseLeft = useCallback<ExpandOrCollapser>(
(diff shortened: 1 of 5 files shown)


src/pivot-table/hooks/use-data-model.ts
```diff
@@ -54,15 +56,16 @@ export default function useDataModel({ model, nextPageHandler, pageInfo }: UseDa
         };
 
         const [pivotPage] = await genericObjectModel.getHyperCubePivotData(Q_PATH, [nextArea]);
-        nextPageHandler(pivotPage);
 
-        return true;
+        // Guard against page changes
+        if (currentPage.current === pageInfo.page) {
+          nextPageHandler(pivotPage);
+        }
```
a-m-dev, line 63: if we change page, since `usePagination` is in supernova level -> it will trigger a completely new render. That means `useDataModel` hook will get triggered from scratch but with new arguments (which happen to has new `pageInfo.page` value)

This means the if check here will always going to be true in case of page change? (or maybe i'm missing something)

cbt1: > if we change page, since usePagination is in supernova level -> it will trigger a completely new render.

Yes

> That means useDataModel hook will get triggered from scratch but with new arguments (which happen to has new pageInfo.page value)

Yes

> This means the if check here will always going to be true in case of page change? (or maybe i'm missing something)

I think if it like this: when new page is rendered, the "old" fetch is still doing it's work and once completed, it stil holds a reference to the old `pageInfo` object. So the `pageInfo` is an old reference bound when creating the function. Even if the the callback function been re-created from scratch.

While `currentPage` will always point to the actual current page not some old reference.

a-m-dev: this explanation makes sense based on the bounded context that holds the prev page info, while it's still currently doing it's job in call stack. 

but one suggestion here would be using [AbortController() api](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) around the fetch request and in case of page changes while still request is in progress, just simply cancel the fetch call. 

That way we have a controlled fetch which would prevent this behaviour, but I'm not sure how could we integrate it with our rpc call to engine. 

but it's not end of the world so, let's get this one in then we could figure it out later if it was possible.

a-m-dev (approved):

src/pivot-table/hooks/use-data-model.ts
cbt1: AbortController would be nice but as far as I know you cannot use that to abort a Promise. It's for `Fetch` only.

PR #388, opened 2023-09-29
fix: use measure text from util

cbt1:
Baseline images updated because the width of the columns have changed. As expected when going from "M" to "N" as measuring character.

The column width is expected to have change with -12% (using default fonts) AND decrease by 25px as the "leeway" is no longer added.


test/rendering/render.mjs-snapshots/scenario-1-chromium-linux.png
cbt1, line 1: I think some logic for measure the label width of measure could be added for "FitToContent" and "Auto". Which would improve the size of the psuedo dimension column.

test/rendering/render.mjs-snapshots/scenario-total-styling-below-chromium-linux.png
cbt1, line 1: The totals label is not included in the "FitToContent" and "Auto" measurement. This becomes rather obvious here. Should considerer including it as the width of that can be determined by simply measuring it.

test/rendering/render.mjs-snapshots/scenario-1-chromium-linux.png
cbt1: So it does measure it by does not take the padding into account so the text ends up overflowing

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -94,14 +99,14 @@ export default function useColumnWidth(
         // Use the max width of all measures
         width = Math.max(
           ...qMeasureInfo.map(({ qFallbackTitle, columnWidth }) => {
-            const fitToContentWidth = measureTextForContent(qFallbackTitle);
+            const fitToContentWidth = measureTextForContent(qFallbackTitle) + TOTAL_CELL_PADDING;
```
cbt1, line 102: The text would not fit if the cell padding was not added

haxxmaxx: Is it only the total label that doesn't fit or is it a general problem?

cbt1: General problem since width of the cell is equal to the width of the text. But the cell has padding so the text it would not fit.

a-m-dev (approved):
LGTM, 
I saw this one: https://github.com/qlik-oss/sn-pivot-table/pull/388/files#diff-bbb4ec4ef3780d1bd2043be54d02d7240fca5be4ca3e830f6934ac8562cc252b

but I think I should take care of it in my PR, seems like a padding issue

PR #389, opened 2023-09-29
fix: include icon width in fitToContent calculation

haxxmaxx:


src/types/types.ts
```diff
@@ -163,6 +163,7 @@ export interface LayoutService {
   hasLeftDimensions: boolean;
   showTotalsAbove: boolean;
   hasPseudoDimOnLeft: boolean;
+  isAlwaysFullyExpanded: boolean;
```
cbt1, line 166: I think you can call this `isFullyExpanded`, since the "always" part seems more like a way of telling Engine that it should be fully expanded.

src/pivot-table/hooks/use-measure-text.ts
```diff
@@ -13,7 +13,7 @@ export interface MeasureTextHook {
   measureText: (text: string) => number;
 }
 
-const MAGIC_DEFAULT_CHAR = "M";
+const MAGIC_DEFAULT_CHAR = "N";
```
cbt1, line 16: This will change with https://github.com/qlik-oss/sn-pivot-table/pull/388

src/types/types.ts
haxxmaxx: makes sense, I will change

src/pivot-table/hooks/use-measure-text.ts
haxxmaxx: Ah, think I changed it to test something, will revert

src/pivot-table/hooks/__tests__/use-column-width.test.ts
```diff
@@ -247,10 +250,24 @@ describe("useColumnWidth", () => {
...
+      layoutService.layout.qHyperCube.qEffectiveInterColumnSortOrder = [0, 1, -1, 2];
 
       const { getRightGridColumnWidth } = renderUseColumnWidth();
       expect(getRightGridColumnWidth()).toBe(width);
     });
+
+    test("should return right column width for non-pseudo dimension for fit to content when dimension is collapsed", () => {
+      const width = 40;
+      mockEstimateWidth(width);
+      mockMeasureText(width);
```
cbt1, line 262: Consider changing the return value of this to `width + 1` so that the test is guaranteed to pass only when `estimatedWidth + EXPAND_ICON_WIDTH` matches

src/pivot-table/hooks/__tests__/use-column-width.test.ts
```diff
@@ -181,6 +182,7 @@ describe("useColumnWidth", () => {
       expect(getRightGridColumnWidth(1)).toBe(ColumnWidthValues.AutoMin);
       expect(getRightGridColumnWidth(2)).toBe(ColumnWidthValues.AutoMin);
     });
+
```
cbt1, line 185: Seems like the Left grid is missing a test case where the icon width is included?

cbt1 (approved):
LGTM, just some minor comments

src/pivot-table/hooks/__tests__/use-column-width.test.ts
haxxmaxx: They are tested, it was already taking icon into account for the left side, before this pr

src/pivot-table/hooks/__tests__/use-column-width.test.ts
haxxmaxx: you mean that I mock with `width` for both of them?

cbt1: If someone was to refactor or change the code from:
```ts
        const fitToContentWidth = Math.max(
          measureTextForHeader(qFallbackTitle),
          estimateWidthForRowContent(qApprMaxGlyphCount) + iconWidth,
        );
```

to:
```ts
        const fitToContentWidth = Math.max(
          measureTextForHeader(qFallbackTitle) + iconWidth,
          estimateWidthForRowContent(qApprMaxGlyphCount),
        );
```

The test would still pass. Which may or may not be intentional.

haxxmaxx: yeah no I don't think that should pass, especially if you change it unintentionally

PR #393, opened 2023-10-02
feat: make auto dependent on other columns

haxxmaxx:
Using the same approach for auto setting as for sn-table. Calculating the independent widths (pixels, percentage, fit to content), subtracting the total width of them, then dividing the remaining width between the auto columns. This is only applicable when the psuedo dimension is deciding the width.

Now also calculating the width at the same time, meaning they are stored in an array. This means we don't have to memoize the function anymore


src/pivot-table/hooks/use-column-width.ts
```diff
@@ -142,89 +137,105 @@ export default function useColumnWidth(
...
+        : averageLeafWidth,
+    [topGridLeavesIsPseudo, leafWidths, layoutService, averageLeafWidth],
   );
 
   // The width of the sum of all columns, can be smaller or greater than what fits in the chart
-  const rightGridFullWidth = useMemo(
-    () => layoutService.size.x * averageLeafWidth,
-    [averageLeafWidth, layoutService.size.x],
-  );
+  const rightGridFullWidth = useMemo(() => size.x * averageLeafWidth, [averageLeafWidth, size.x]);
```
cbt1, line 238: Since you're changing this line. You might as well remove the `useMemo`

cbt1 (approved):
Failing test but LGTM

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -142,89 +137,105 @@ export default function useColumnWidth(
...
-        case ColumnWidthType.Auto: {
-          // TODO: we might need to redo this in the pseudo dimension case, since if not all measure are set to auto
-          // you don't fill upp the width anyway
-          specifiedWidth = autoWidth;
-          break;
+        const addKnownWidth = (width: number) => {
+          widths[idx] = Math.min(ColumnWidthValues.PixelsMax, Math.max(ColumnWidthValues.PixelsMin, width));
+          // remove the width * number of instances of that width from the remaining width for auto columns
+          sumAutoWidths -= widths[idx] * (size.x / columnArray.length);
+        };
```
a-m-dev, line 162: you can put this out of iteration, the only value it needs is the idx which you can pass to it

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -142,89 +137,105 @@ export default function useColumnWidth(
...
-      estimateWidthForContent,
-      measureTextForColumnContent,
-      qMeasureInfo,
-      estimateWidthForColumnContent,
-      leavesIconWidth,
-    ],
-  );
+    if (autoColumnIndexes.length) {
+      // divides remaining width evenly between auto columns
+      const totalNumberOfAutoColumns = (size.x * autoColumnIndexes.length) / columnArray.length;
```
a-m-dev, line 197: shouldn't `totalNumberOfAutoColumns` be equal to just `autoColumnIndexes.length`? 
otherwise the name is misleading in this case, because what this variable store is something else

haxxmaxx: yeah naming could probably be better here. The thing is that we loop over the number of measure, let's say there are 3 => `columnArray.length === 3`. Then one of them is  non-auto => `autoColumnIndexes.length === 2`. Now there could be multiple instances of each measure, if you have a dimension above it for example, let's say a dimension with 10 values => `size.x === 30`. So what `totalNumberOfAutoColumns` is, is the actual number of columns that will have the auto width => 30 * 2 / 3 = 20.

So maybe the name could be `autoColumnInstances` or something?

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: sure

src/pivot-table/hooks/use-column-width.ts
a-m-dev: Now i understand the logic, right!
Yes, the new name is great. 
Also by that, it would be nice to change `autoWidth` (in the line below) to something like `possibleAutoColWidth` or `tunedAutoColWidth` or something better if you have in mind. 

Also can you add this example in code as well, I think it would really help us understand the code better later on when we come back to code base

a-m-dev (approved):

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: sure I can comment something

haxxmaxx: I think `autoWidth` is good, since that is the width every auto column would get (with the exception of when it is lower than the min value)

PR #394, opened 2023-10-02
docs: add columnWidth to initial properties

haxxmaxx:


src/qae/initial-properties.js
```diff
@@ -74,6 +74,43 @@ const properties = {
...
+ * @typedef {object} InlineMeasureDef
+ * @extends NxInlineMeasureDef
+ * @property {ColumnWidth=} columnWidth
+ */
+
+/**
+ * Column width info. For the left grid, the properties are always applied.
+ * For the right grid, only the leaf nodes will listen to the properties, and the columns above will get the width of the leaves accumulated
+ * @typedef {object} ColumnWidth
+ * @property {('auto' | 'FitToContent' | 'pixels' | 'percentage')} type - Defines how the column width is set. For the right grid, `auto` calculates the width(s) so the total width of the columns equals the right grid width. If the width reaches a minimum value, the columns will overflow. For the left grid, `auto` is N/A and defaults to `fitToContent`. `fitToContent` calculates a width based on the cells' content. `pixels` uses a specified pixel value. `percentage` sets the column width to specified percentage of the chart width
```
haxxmaxx, line 109: Trying to make this concise, accurate and easy to understand. Feedback appreciated 🙃

cbt1, line 109: `cells'` should maybe be `cells`

a-m-dev: I have a question about percentage, what if the sum of values that user sets reaches 100%. I heared once that this 100% is the same as `rect.width`, considering the 75% rule on left grid, if i have 5 cols in left grid and set 20% for each, how is that going to shown to the user?

haxxmaxx: well, they would shrink to 15% each. I guess I based this description on what it will work as when it is released. Then we will be able to scroll, and 20% * 5 will result in the same width as the chart, just that you have to scroll to see everything

haxxmaxx: no but it should be `cell's`, good catch

a-m-dev (approved):

src/qae/initial-properties.js
haxxmaxx: actually it should probably be `column's`

cbt1:
Why the API spec error?

cbt1:
> > Why the API spec error?
> 
> It warns that it can't find `NxInlineDimensionDef` and `NxInlineMeasureDef`. But then it throws an Cannot read properties of undefined error. They are used in sn-table so don't see why it wouldn't find the type here

Should figure that out before mergin this to main then I'd say

PR #396, opened 2023-10-03
fix: prevent clickable head cell for pseudo dims

a-m-dev:
fix for this: 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/efd8db39-79cb-4f8b-b0cc-1643b80b052a

### DIFF
--- a/src/pivot-table/components/cells/DimensionTitleCell.tsx
+++ b/src/pivot-table/components/cells/DimensionTitleCell.tsx
@@ -93,7 +93,6 @@ const DimensionTitleCell = ({
       style={{
         ...style,
         ...getBorderStyle(true, isLastColumn, styleService.grid.border),
-
         color,
       }}
       data-testid={testId}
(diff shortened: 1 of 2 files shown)


src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -89,6 +89,7 @@ const DimensionTitleCell = ({
       background={open ? activeBackground : background}
       hoverBackground={hoverBackground}
       shouldShowMenuIcon={shouldShowMenuIcon}
+      isDimension={isDim}
```
cbt1, line 92: Set pointerEvents here instead? No need for the logic in `StyledHeaderCellWrapper`?

```suggestion
      pointerEvents={isDimension ? "all" : "none"}
```

a-m-dev: Yes, thought of it, but it's not a good practice IMO, since we have all styles in `StyledHeaderCellWrapper`

cbt1 (approved):

PR #397, opened 2023-10-04
feat: separate scrolls

a-m-dev:
this PR is ok for column values as :
- [x] Pixels
- [x] Percentage 
- [x] fit to content
- [x] auto
- [x] mix of those 👆


src/pivot-table/components/shared-styles.ts
```diff
@@ -90,3 +91,27 @@ export const getTotalCellDividerStyle = ({
...
 };
+
+interface StyledLeftGridProps {
+  leftGridWidth: number;
+}
+
+export const StyledLeftGrid = styled("div", {
+  shouldForwardProp: (prop: string) => !["leftGridWidth"].includes(prop),
+})(({ leftGridWidth }: StyledLeftGridProps) => ({
+  width: leftGridWidth,
```
cbt1, line 102: Not sure this is a great idea. I think that every time the chart is re-size or a column is re-size a new css class will be created by mui

cbt1, line 102: Not sure this is a great idea. I think that every time the chart is re-size or a column is re-size a new css class will be created by mui

a-m-dev: I interpret that as a new (re)render cycle and every thing will start from scratch which is already happening for/in other (styled) components as well, do you see any problem with that?

cbt1: When testing re-sizing the chart, it generated a new inline style sheets on every re-size event. I needed to refresh the browser for the style sheets to get purged.

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -28,6 +28,11 @@ export interface GetHeaderCellsIconsVisibilityStatus {
   };
 }
 
+interface LeftGridColumnWidths {
+  widths: number[];
+  originalWidths: number[];
```
cbt1, line 33: I found `originalWidths` to be hard to figure out why it was needed. Before I realized that it's there to get the full width of the columns. Maybe rename it to `fullWidths`? As the `widths` is "capped" at the 75% rule right?

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -119,10 +124,13 @@ export default function useColumnWidth(
     });
 
     const leftGridMaxWidth = rect.width * LEFT_GRID_MAX_WIDTH_RATIO;
-    if (sumOfWidths < leftGridMaxWidth) return widths;
+    if (sumOfWidths < leftGridMaxWidth) return { widths, originalWidths: widths };
 
     const multiplier = leftGridMaxWidth / sumOfWidths;
-    return widths.map((w) => w * multiplier);
+    return {
```
cbt1, line 130: I think you can refactor this so that it returns both the `fullWidths` array and the value for `leftGridWidth`. That would mean that you don't need the `leftGridWidth` on line 179 and only need to keep one array of widths.

src/pivot-table/hooks/use-column-width.ts
a-m-dev: I thought of this a bit more and came up with more clearer names, now we have:
`actualColumnWidths`: which refers to the base/original width per column that we receive from sense-client 
`modifiedColumnWidth`:  which reflects the widths per column if we modify its value (currently only being used to calculate the left grid width)

Hope it sounds well

src/pivot-table/hooks/use-column-width.ts
a-m-dev: nice point, actually I can calculate it once, independent of return or if statement (above them) and then only return the result in each case. (of course multiplied by .75 coefficient)

+ 

i can merge `leftGridWidth` (as you mentioned) and `getLeftGridColumnWidth` also with the return object and access those from there. a minor refactor per se!

a-m-dev: Maybe a dumb question but worth asking:
why do we need to loop over `widths` array to multiply each value by ratio we have? 
```
widths.reduce((totalWidth, w) => totalWidth + w * multiplier, 0)
```

why not just not multiply `sumOfWidths` we already have with ratio? or in fact we already have it there -> `leftGridMaxWidth` is just sitting there in front of us.

cbt1: Nice catch!

cbt1: Are you going to make another change here? You are currently returning two arrays of column widths

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -166,7 +179,7 @@ export default function useColumnWidth(
   );
 
   const leftGridWidth = useMemo(
-    () => leftGridColumnWidths.reduce((totalWidth, w) => totalWidth + w, 0),
+    () => leftGridColumnWidths.modifiedColumnWidth.reduce((totalWidth, w) => totalWidth + w, 0),
```
haxxmaxx, line 182: If I understand what you are trying to do, the only changes you need is to scrap the whole scaling logic in `leftGridColumnWidths` and only return one array as before. And then here you do
```suggestion
    () => Math.min(rect.width * LEFT_GRID_MAX_WIDTH_RATIO, leftGridColumnWidths.reduce((totalWidth, w) => totalWidth + w, 0)),
```

src/pivot-table/hooks/use-column-width.ts
a-m-dev: Yes, I was thinking in addition of `actualColumnWidths` and `modifiedColumnWidths`, we can include `leftGridWidth` and `getLeftGridColumnWidth(idx)` into that useMemo since all 4 is related to the left grid, I'll update this PR soon

src/pivot-table/hooks/use-column-width.ts
cbt1: Seems like that would work to me

a-m-dev: I actually want to escape runtime/more calculations on each call to `leftGridWidth`,
by latest changes, it's calculated once within the `leftGridWidthInfo` and `useMemo`-ised

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: I like the approach to calculate leftGridWidth in here, but you still don't need modifiedColumnWidths. if you do the `Math.min` I suggested below, you can skip all the logic from line 135 as well.

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -90,9 +95,9 @@ export default function useColumnWidth(
       }
     };
 
-    let sumOfWidths = 0;
+    let sumOfActualWidths = 0;
 
-    const widths = visibleLeftDimensionInfo.map((qDimensionInfo, index) => {
+    const actualColumnWidths = visibleLeftDimensionInfo.map((qDimensionInfo, index) => {
```
haxxmaxx, line 100: the actual isn't really needed imo, since there is only one set of widths

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -118,11 +123,11 @@ export default function useColumnWidth(
...
 
-    const leftGridMaxWidth = rect.width * LEFT_GRID_MAX_WIDTH_RATIO;
-    if (sumOfWidths < leftGridMaxWidth) return widths;
-
-    const multiplier = leftGridMaxWidth / sumOfWidths;
-    return widths.map((w) => w * multiplier);
+    return {
+      columnWidths,
+      leftGridWidth: Math.min(rect.width * LEFT_GRID_MAX_WIDTH_RATIO, sumOfWidths),
+      getLeftGridColumnWidth: (idx) => columnWidths[idx],
```
cbt1, line 129: I don't really see any point in having this function anymore. Might as well just consume the `columnWidths` array. But no need to change it in this PR.

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -287,18 +288,17 @@ export default function useColumnWidth(
...
-    [leftGridWidth, rightGridFullWidth],
+    () => leftGridWidthInfo.leftGridWidth + rightGridFullWidth + GRID_BORDER,
+    [leftGridWidthInfo.leftGridWidth, rightGridFullWidth],
   );
 
   const showLastRightBorder = useMemo(() => totalWidth < rect.width, [totalWidth, rect.width]);
 
   return {
-    leftGridWidth,
+    leftGridWidthInfo,
```
haxxmaxx, line 298: I think I'd prefer same return structure as before. So we asign `leftGridWidth` and `getLeftGridColumnWIdth` here, and don't have to do the destructuring in PivotTable

haxxmaxx (approved):
minor comments, feel free to merge. I just saw Christofers comment, we could do something about what we expose from the hook in a different PR

src/pivot-table/hooks/use-column-width.ts
a-m-dev: I removed it

src/pivot-table/hooks/use-column-width.ts
a-m-dev: i can spread the result of leftGridWidthInfo in here and the hooks interface will extends that one like: 
```
interface ColumnWidthHook extends LeftGridWidthInfo { ... }
```

this will get rid of destructuring in pivot table!

haxxmaxx: sounds good

PR #399, opened 2023-10-06
feat: header font styling

cbt1:
Adds an option to the styling panel to set the header text "bold", "italic" and "underline". Currently there is no support for this in the theme. The options can only be set via the styling panel.

TODO in a separate PR:
- Measure text using bold and italic when applicable

Header cell:
Skärmavbild 2023-10-06 kl  10 25 41

Styling panel options to set "bold", "italic" and "underline":
Skärmavbild 2023-10-06 kl  10 25 34


src/types/types.ts
```diff
@@ -222,20 +225,21 @@ interface ColumnTitleStyling extends CellStyling {
   activeBackground: string;
 }
 
-interface HeaderStyling extends Pick<FontStyling, "fontSize" | "fontFamily"> {
+interface HeaderStyling
+  extends Pick<FontStyling, "fontSize" | "fontFamily" | "fontWeight" | "fontStyle" | "textDecoration"> {
```
a-m-dev, line 229: In This case, Omit<FontStyling, "color"> might be helpful

src/types/types.ts
```diff
@@ -201,9 +201,12 @@ export interface PageInfo {
   rowsOnCurrentPage: number;
 }
 
-interface FontStyling {
+export interface FontStyling {
   fontSize: string;
   fontFamily: string;
+  fontWeight: "600" | "normal";
+  fontStyle: "italic" | "normal";
+  textDecoration: "underline" | "none";
```
a-m-dev, line 209: maybe overkill but, you can have a basicFontStyling interface that this one extends and use it below instead of pick

src/services/style-service.ts
```diff
@@ -63,6 +65,20 @@ const resolveColor = (theme: ExtendedTheme, color: PaletteColor | undefined) =>
 const fontSizeToCellHeight = (fontSize: string, lineClamp: number) =>
   +(parseInt(fontSize, 10) * LINE_HEIGHT_COEFFICIENT * lineClamp + CELL_PADDING_HEIGHT).toFixed(2);
 
+const resolveFontWeight = (fontStyleOptions: FontStyleOptions[] | undefined) => {
+  if (fontStyleOptions === undefined) {
+    return BOLD_FONT_WEIGHT;
+  }
```
a-m-dev, line 71: does this means the default font weight is bold for all headers?

a-m-dev (approved):
LGTM, just minor comments

src/services/style-service.ts
cbt1: Yes, by default `fontStyleOptions` is `undefined`

src/types/types.ts
cbt1: I'll wait and see how it plays out when I add all these styling options for some of the others sections as well

PR #401, opened 2023-10-09
fix: removed postinstall

enell:


cbt1 (approved):

cbt1:
Might need to update the shared workflow to always run `locale:generate`

PR #403, opened 2023-10-10
Crl/drag to resize

haxxmaxx:


src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -66,7 +70,11 @@ export default function useSelectionsModel(
   );
 
   const select = useCallback(
-    (qType: EngineAPI.NxSelectionCellType, qRow: number, qCol: number) => async () => {
+    (qType: EngineAPI.NxSelectionCellType, qRow: number, qCol: number) => async (evt: React.MouseEvent) => {
+      if ((evt.target as HTMLElement)?.className.includes("sn-pivot-table-column-adjuster")) {
```
haxxmaxx, line 74: Not sure if this is the place to check for this, but need to cancel selection if the event comes from the adjuster

src/pivot-table/data/helpers/create-cell.ts
```diff
@@ -83,6 +84,7 @@ const createCell = (
     isEmpty: node.qType === NxDimCellType.NX_DIM_CELL_EMPTY,
     isNull: node.qType === NxDimCellType.NX_DIM_CELL_NULL,
     isPseudoDimension: node.qType === NxDimCellType.NX_DIM_CELL_PSEUDO,
+    isPseudoDimensionBefore: !!(parent?.isPseudoDimensionBefore || parent?.isPseudoDimension),
```
haxxmaxx, line 87: This information is kind of already on the cell. You can search up the parents and see if anyone is pseudo, but this skipes that loop

src/pivot-table/components/cells/utils/should-render-column-adjuster.ts
```diff
@@ -0,0 +1,5 @@
+import type { Cell } from "../../../../types/types";
+
+export default function shouldRenderColumnAdjuster(cell: Cell, isActive: boolean) {
```
haxxmaxx, line 3: went back and forth on this. I'd prefer if ColumnAdjuster took care of its own render condition. The thing is if we add <ColumnAdjuster> to all dimension and pseudo dimension cells, there is a useEffect that runs on all of them, so I went with just not rendering it at all. The logic for rendering it will become more complex when it is added to the header grid

src/pivot-table/components/cells/styles.tsx
```diff
@@ -68,3 +68,34 @@ export const StyledLabel = styled("div", {
...
+  height: "100%",
+  top: 0,
+  left: `100%`,
+  cursor: "col-resize",
+  // last column padding, other double padding + border
+  width: `${isLastColumn ? 4 : 9}px`,
+  justifyContent: isLastColumn ? "flex-end" : "center",
+  "&&:hover:not(:focus, :active)": {
+    "& .sn-pivot-table-column-adjuster-border": {
+      background: "#D9D9D9",
```
cbt1, line 86: `background` probably needs to be a bit more dynamic. As some cell might have this exact color or something other color which will give low contrast.

src/pivot-table/components/cells/styles.tsx
```diff
@@ -68,3 +68,34 @@ export const StyledLabel = styled("div", {
   fontSize,
   fontFamily,
 }));
+
+export const AdjusterHitArea = styled(Box, {
```
cbt1, line 72: Any value in having this being a `Box` instead of just `"div"`?

src/pivot-table/data/helpers/create-cell.ts
cbt1: Consider renaming it to `isAncestorPseudoDimesion` as that seem to be exactly the kind of value it holds?

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Component that is placed on top of column border, to resize the columns.
+ * When you start dragging, mouse move and mouse up listeners are added.
+ * While dragging this components follows the pointer, and on mouse up all column widths are updated.
+ */
+const ColumnAdjuster = ({ cell, columnWidth, dataModel }: AdjusterProps) => {
+  const [internalWidth, setInternalWidth] = useState(columnWidth);
+  const tempWidth = useRef({ initWidth: 0, columnWidth: 0, initX: 0 });
+  const LEFT_ADJUSTMENT = CELL_PADDING + GRID_BORDER;
```
cbt1, line 24: This can be declared outside the component

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  const handleDoubleClick = () => dataModel?.applyColumnWidth({ type: ColumnWidthType.FitToContent }, cell);
+
+  return (
+    <AdjusterHitArea
+      style={{ left: internalWidth - LEFT_ADJUSTMENT }}
+      isLastColumn={false}
+      className="sn-pivot-table-column-adjuster"
+      key={`adjuster-${cell.x}`}
```
cbt1, line 67: Would think this will cause duplicate keys when you also have the component in the "LeftGrid"

cbt1:
FYI: Dragging in last column when all columns dont fill up the whole chart does not render the "ColumnAdjuster"

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/51bcfc1d-0873-4e35-b49e-90ca0a565c22

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -0,0 +1,76 @@
...
+      initX: evt.clientX,
+      initWidth: columnWidth,
+      columnWidth,
+    };
+
+    document.addEventListener("mousemove", mouseMoveHandler);
+    document.addEventListener("mouseup", mouseUpHandler);
+  };
+
+  const handleDoubleClick = () => dataModel?.applyColumnWidth({ type: ColumnWidthType.FitToContent }, cell);
```
cbt1, line 60: Love this one, feels really nice to use ❤️‍🔥

cbt1:
Minor stuff but the "Adjuster" is "blinking/flashing" a bit when you drag it fast. Also the "Adjuster" disappear if you move outside the row.

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/12302005-99dd-4953-8800-169f42b789a1

src/pivot-table/components/cells/ColumnAdjuster.tsx
haxxmaxx: yes, there is already a key warning now when you mention it, I should look at that

src/pivot-table/components/cells/styles.tsx
haxxmaxx: nope, just copy pasta from sn-table. Maybe should be div for consistency in the repo

src/pivot-table/components/cells/styles.tsx
haxxmaxx: true, you have logic for that somewhere right?

cbt1: Thinking some more about it. When the dimension has background color by expression, the color can be different for each cell. That case is probably not worth trying to handling, but if the cell is getting its color from the `StyleService`, it's going to be the same for all cells. In that case case you could use the `getHoverColor()` from the utils repo to get a contrasting color.

cbt1: But I'd say lets handle that special case in a different PR

src/pivot-table/data/helpers/create-cell.ts
haxxmaxx: Ah was looking for the correct term, thanks

src/pivot-table/components/cells/ColumnAdjuster.tsx
haxxmaxx: Thank microsoft for inventing it and edvard for implementing it in the old table I guess ;)

src/pivot-table/components/cells/styles.tsx
haxxmaxx: yeah exactly, this component need to contrast two different backgrounds potentially. I'll write it down

src/pivot-table/components/cells/ColumnAdjuster.tsx
haxxmaxx: I don't see that warning anymore. I'll wait with adjusting the key when I add it to the header grid.

haxxmaxx:
> Minor stuff but the "Adjuster" is "blinking/flashing" a bit when you drag it fast. Also the "Adjuster" disappear if you move outside the row.

the flickering is there in sn-table as well. Would be nice if we can get rid of it. Right now we control the cursor by setting the css for hover over the border. When the border doesn't keep up, you get the other pointer. I can look at it but not sure how to change it. or just optimize it so there is no lag...

The dissaperence isnt great, I'll look at that too

src/pivot-table/hooks/use-data-model.ts
```diff
@@ -68,15 +82,53 @@ export default function useDataModel({ model, nextPageHandler, pageInfo }: UseDa
...
+          }
+        : {
+            qPath,
+            qOp: "Add" as EngineAPI.NxPatchOpType,
+            qValue: JSON.stringify(newColumnWidth),
+          };
+
+      // typescript doesn't like unresolved promises, so we have to do a no-op then
+      // there is nothing that needs to happen after this, so no need for the function to be async
+      model?.applyPatches([patch], true).then(
```
cbt1, line 114: I think you can do `await model?.applyPatches([patch], true);` instead

haxxmaxx: That makes this function async, then the function we use for mouseup becomes async, and then I get a type error

cbt1: Odd I did not get the type error when I tested it. Did you get the error in VSC and/or from `yarn types:check`

haxxmaxx: good question, I can try it again

haxxmaxx: it is a lint issue. I could just ignore it on this line then no function would be async

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Component that is placed on top of column border, to resize the columns.
+ * When you start dragging, mouse move and mouse up listeners are added.
+ * While dragging this components follows the pointer, and on mouse up all column widths are updated.
+ */
+const ColumnAdjuster = ({ cell, columnWidth, dataModel }: AdjusterProps) => {
+  const [internalWidth, setInternalWidth] = useState(columnWidth);
+  const tempWidth = useRef({ initWidth: 0, columnWidth: 0, initX: 0 });
+
+  useEffect(() => setInternalWidth(columnWidth), [columnWidth]);
```
cbt1, line 27: This seem like a good use case for `useOnPropsChange` hook (https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)

src/pivot-table/components/cells/styles.tsx
```diff
@@ -67,3 +67,34 @@ export const StyledLabel = styled("div", {
...
 }));
+
+export const AdjusterHitArea = styled("div", {
+  shouldForwardProp: (prop: string) => prop !== "isLastColumn",
+})(({ isLastColumn = false }: { isLastColumn: boolean }) => ({
+  display: "flex",
+  position: "absolute",
+  height: "100%",
+  top: 0,
+  left: `100%`,
```
cbt1, line 78:
```suggestion
  left: "100%",
```

cbt1 (approved):
LGTM, just some final minor comments

src/pivot-table/components/cells/styles.tsx
haxxmaxx: it shouldn't be ther at all, we set left by inline styling, since it determines the positon

src/pivot-table/components/cells/ColumnAdjuster.tsx
haxxmaxx: oh nice

PR #404, opened 2023-10-11
feat: new styling panel design

cbt1:
Major areas updated in this PR:
- Styling panel `src/ext/property-panel/settings/styling-panel`
- Styling service `src/services/style-service.ts`
- All consumers of the style service (mostly cell components and useColumnWidth hook)
- Types have change a lot. Specifically the `StyleService` and `Component` types. Where `Component` change is breakning from the perspective of object properties (but not part of API spec yet).

TODO after this PR:
- Update measure text to include bold/italic where applicable
- Investigate why italic text can get clipped in the cell (Firefox)
- Update Sense themes with new structure
- Add styling options to API spec

New styling panel:
Skärmavbild 2023-10-11 kl  10 42 49

New styling pivot table:
Skärmavbild 2023-10-11 kl  10 42 58


src/types/types.ts
```diff
@@ -201,58 +201,36 @@ export interface PageInfo {
...
+};
 
-export interface StylingOptions {
-  header: HeaderStyling;
-  content: MeasureContentStyling;
-  rowContent: DimensionContentStyling;
-  columnContent: DimensionContentStyling;
-  grid: GridStyling;
-}
+export type ThemeStyling = {
```
cbt1, line 214: Using `type` instead to imrpove the info you get in VSC when hovering the type

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -102,7 +101,7 @@ export default function useColumnWidth(
         // Use the max width of all measures
         width = Math.max(
           ...qMeasureInfo.map(({ qFallbackTitle, columnWidth }) => {
-            const fitToContentWidth = measureTextForContent(qFallbackTitle) + TOTAL_CELL_PADDING;
+            const fitToContentWidth = measureTextForMeasureValue(qFallbackTitle) + TOTAL_CELL_PADDING;
```
haxxmaxx, line 104: hmm looks like this was wrong before. This should measure the width of the text for the measure labels, so I guess this should be `measureTextForDimensionValue` now?

src/ext/property-panel/settings/styling-panel/index.ts
```diff
@@ -16,10 +16,18 @@ const getStylingPanelConfig = (translator: stardust.Translator) => ({
...
+          section: "dimensionValues",
+          translator,
+        }),
+        measureValueSection: largePanelSection({
+          section: "measureValues",
+          translator,
+        }),
+        measureLabelSection: smallPanelSection({ section: "measureLabels" }),
+        totalValuesSection: smallPanelSection({ section: "totalValues", defaultFontStyle: ["bold"] }),
+        nullValuesSection: smallPanelSection({ section: "nullValues", fallbackBackground: Colors.Black5 }),
```
haxxmaxx, line 30: Very nice to generalize it in this way

src/pivot-table/components/Disclaimer.tsx
```diff
@@ -11,7 +11,7 @@ interface DisclaimerProps {
 
 const Disclaimer = ({ styleService, translator }: DisclaimerProps): JSX.Element => {
   const text = translator.get("SNPivotTable.LimitedData");
-  const { fontFamily } = styleService.content; // TODO Resolve from root of theme?
+  const { fontFamily } = styleService.measureValues; // TODO Resolve from root of theme?
```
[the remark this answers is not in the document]
cbt1: Agree, will change to base theme font and fallback to the default hard-coded font if needed

src/pivot-table/hooks/use-column-width.ts
cbt1: Nice catch, will fix

haxxmaxx (approved):

PR #405, opened 2023-10-11
fix: double scroll

a-m-dev:
**including in this PR:** 
- [x] main fix for removing double scrollbar 
- [x] separating left grid container and data grid container -> which handles the x-axis scroll
- [x] introduce a wrapper scrollable container -> which handles the y-axis  scroll
- [x] attach scrollbar to the bottom of scrollable area when there are less rows than `rect.height`

**NOT included in this PR:**
- [x] scrollbar overlay (vertically/horizontally) on data grid (https://github.com/qlik-oss/sn-pivot-table/pull/417)

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/72bd3d6f-5474-46d6-ba6e-976e4e91b112


src/pivot-table/components/PivotTable.tsx
```diff
@@ -82,68 +99,113 @@ export const StickyPivotTable = ({
...
-                height={topGridHeight}
-                headersData={headersData}
-                translator={translator}
-                changeSortOrder={changeSortOrder}
-                changeActivelySortedHeader={changeActivelySortedHeader}
-              />
+            <ScrollableContainer
+              ref={horizontalScrollableContainerRef}
+              width={leftGridWidth}
+              height={rowsCanFitInTableViewPort ? topGridHeight + dataGridHeight + GRID_BORDER : tableRect.height}
```
cbt1, line 126: Seems like this is use in two places. Maybe move it to a const?

cbt1:
I've played around with this and it seem to be working nicely. Why the Draft status?

a-m-dev:
@cbt1, Mainly because of two special edge cases where scrollbar gets over data row, 

**in case of Vertical scrollbar:** 
Screenshot 2023-10-13 at 09 35 43


**in case of Horizontal scrollbar:**
Screenshot 2023-10-13 at 09 36 37

a-m-dev:
So far i fixed: 
1. vertical scrollbar overlay on data grid
2. horizontal scrollbar overlay on data grid (when page is full of data and scrollable)
The only thing left is this one here: 
274829900-deaba513-1a6c-4250-8132-7f3fe9bb13cb

I'll give it a try and update this pr, and keep going on this one and after i've done all work and cleaned up every thing (lint, test, refactors, last touches...) I can separate each in their own PR (since i have commits for each) simply by cherry picking  them.

a-m-dev:
As I try to fix edge cases, the code becomes more complex, probably @cbt1 you are right about separating PR's. 
will prepare this one (without overlay problems) and then work on that problem in separate PR

src/pivot-table/components/PivotTable.tsx
a-m-dev: Yes, and this one will be addressed in next PR (scrollbar overlays on data)

src/pivot-table/hooks/use-grid-height.ts
```diff
@@ -24,5 +24,15 @@ export default function useGridHeight({ pageInfo, headersData, topDimensionData,
   const rowsCanFitInTableViewPort = Math.floor(tableRect.height / contentCellHeight);
   const showLastBottomBorder = pageInfo.rowsOnCurrentPage < rowsCanFitInTableViewPort;
 
-  return { containerHeight, topGridHeight, leftGridHeight, dataGridHeight, showLastBottomBorder };
+  const isEmptySpaceExistsBelowLastRow = topGridHeight + GRID_BORDER + dataGridHeight < tableRect.height;
```
cbt1, line 27: I would think this should be the same value as `showLastBottomBorder`? But `showLastBottomBorder` does appear to be bugged as it does not take `topGridHeight + GRID_BORDER` into account?

src/pivot-table/components/shared-styles.ts
```diff
@@ -94,8 +94,8 @@ export const getTotalCellDividerStyle = ({
 
 export const getLeftGridStyles = (leftGridWidth: number): React.CSSProperties => ({
   width: leftGridWidth,
-  overflow: "scroll",
   position: "relative",
   display: "flex",
   flexDirection: "column",
+  overflow: "visible",
```
cbt1, line 100: I guess this was for testing purpose?

src/pivot-table/components/containers/ScrollableContainer.tsx
```diff
@@ -1,26 +1,40 @@
...
 const ScrollableContainer = (props: ScrollableContainerProps, ref: React.LegacyRef<HTMLDivElement>): JSX.Element => {
-  const { rect, children, onScroll } = props;
+  const { width, height, children, onScroll, style, showVerticalScrollbar, showHorizontalScrollbar, origin } = props;
   const { interactions } = useBaseContext();
 
   return (
     <div
       ref={ref}
-      data-testid="scrollable-container"
+      data-testid={getTestId(origin)}
```
cbt1, line 30: The data attribute should not be named "testid", if it's a value that we use for anything else then testing. Maybe just call it `data-key` instead?

src/pivot-table/components/containers/ScrollableContainer.tsx
```diff
@@ -1,26 +1,40 @@
 import React from "react";
-import type { Rect } from "../../../types/types";
 import { useBaseContext } from "../../contexts/BaseProvider";
 
+export enum ScrollableContainerOrigin {
```
cbt1, line 4: Put this in `src/types/types.ts` since it's used in a lot places? Not only this file and its test file.

src/pivot-table/components/shared-styles.ts
a-m-dev: there is actually no need for those inner (wrapper) divs now (and consequently this style as well) since I changed the children of `<StickyContainer />` to `ReactNode` -> now it can accept multiple children. 

this will be removed

src/pivot-table/hooks/use-grid-height.ts
a-m-dev: fixed it with latest commit, 
but in terms of equality (showLastBottomBorder == isEmptySpaceExistsBelowLastRow) not quite sure if they are representing same value since `showLastBottomBorder` is computed value based on (header/content)rowHeight but `isEmptySpaceExistsBelowLastRow` is the explicit grid heights. so i think `isEmptySpaceExistsBelowLastRow` is much more precise (not Math.floor used for it) + in next PR, i have to consider horizontal scrollbar height into this calculation, so the purpose of them are different as well

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/0babcf5a-3b36-4a7e-b378-58e76faec865

cbt1: `showLastBottomBorder` should be true only if all rows are visible to the user? Here is an example where it fails with that:


https://github.com/qlik-oss/sn-pivot-table/assets/16608020/3b59cc01-c1e9-48b2-bd5d-2238621501c0


I'm not really sure what the purpose of `isEmptySpaceExistsBelowLastRow` is, it it's not the same?

cbt1: I think `isEmptySpaceExistsBelowLastRow` is the only variable we should use. But perhaps reverse the logic and rename it to `areAllRowsVisible`? So `showLastBottomBorder` can be removed since it's broken

cbt1 (approved):
🚢

PR #406, opened 2023-10-13
feat!: new components api

cbt1:
Updates the API to adhere to the new components api defined in #404

### DIFF
--- a/api-specifications/properties.json
+++ b/api-specifications/properties.json
@@ -121,8 +121,8 @@
           "kind": "object"
         },
         "components": {
-          "description": "Holds general styling",
-          "nullable": true,
+          "description": "General and chart specific styling",
+          "optional": true,
           "kind": "array",
           "items": {
             "type": "#/definitions/Component"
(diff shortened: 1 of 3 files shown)


api-specifications/properties.json
```diff
@@ -229,70 +229,229 @@
...
-        "hoverFontColor": {
-          "description": "When only `hoverColor` is set, this is adjusted to either `#f4f4f4` or `#ffffff` for optimal contrast",
+        "background": {
+          "description": "Cell background color palette",
           "optional": true,
           "type": "#/definitions/PaletteColor"
+        }
+      }
+    },
+    "PartialCellStyling": {
```
cbt1, line 411: No idea what to call this. Maybe there is a better way to pick some props from another type?

src/qae/initial-properties.js
```diff
@@ -117,34 +117,98 @@ const properties = {
...
+ */
+
+/**
+ * Mandatory key for chart styling
+ * @name ChartStylingKey
+ * @type {"theme"}
+ */
+
+/**
+ * Custom styling of cells
```
cbt1, line 171: Maybe change to just `Chart styling`?

src/qae/initial-properties.js
```diff
@@ -117,34 +117,98 @@ const properties = {
...
+ * @type object
+ * @property {GeneralStylingKey} key
+ * @property {TitleStyling} title
+ */
+
+/**
+ * Title styling options
+ * @name TitleOptions
+ * @type object
+ * @property {string} [fontSize] - Font size in pixel value
```
haxxmaxx, line 149: I don't recognize this syntax, are the [] to highlight the prop name?

api-specifications/properties.json
haxxmaxx: Yeah you could reuse CellStyling and omit the props not needed. But you still would have to name it right? I think the current name is fine.

haxxmaxx (approved):

api-specifications/properties.json
cbt1: Yes I don't think there is a way around that

src/qae/initial-properties.js
cbt1: `@property {string} [fontSize]` means that `fontSize` is optional. Which is the JSDoc way of doing it. You can also do @property {string=} fontSize which would be the Google Closure compiler syntax, that also works.

src/qae/initial-properties.js
haxxmaxx: yeah I think so, the grid stuff is not really applied to independent cells

src/qae/initial-properties.js
haxxmaxx: ah ok, didn't know that. I think I prefer this way

PR #411, opened 2023-10-16
fix: use font weight in colum width

cbt1:
When measure text to calculate the column widths, the font weight should be included to get correct width.

### DIFF
--- a/src/pivot-table/hooks/use-column-width.ts
+++ b/src/pivot-table/hooks/use-column-width.ts
@@ -2,7 +2,7 @@ import { useMeasureText } from "@qlik/nebula-table-utils/lib/hooks";
 import { useCallback, useMemo } from "react";
 import { PSEUDO_DIMENSION_INDEX } from "../../constants";
 import { ColumnWidthType, type ColumnWidth } from "../../types/QIX";
-import type { LayoutService, Rect, VisibleDimensionInfo } from "../../types/types";
+import type { CellStyling, LayoutService, Rect, VisibleDimensionInfo } from "../../types/types";
 import { CELL_PADDING } from "../components/shared-styles";
 import { GRID_BORDER, HEADER_ICON_SIZE } from "../constants";
 import { useStyleContext } from "../contexts/StyleProvider";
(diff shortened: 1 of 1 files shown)


src/pivot-table/hooks/use-column-width.ts
```diff
@@ -67,12 +69,16 @@ export default function useColumnWidth(
   const styleService = useStyleContext();
   const { measureText: measureTextForHeader } = useMeasureText({
     ...styleService.header,
-    bold: true,
+    bold: isBold(styleService.header),
+  });
+  const { estimateWidth: estimateWidthForMeasureValue } = useMeasureText({
```
cbt1, line 74: TODO: Update `useMeasureText` to except `fontWeight` as a property instead. This would be a "breaking change" in the utils repo and require updates in both tables repos.

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -45,6 +45,8 @@ export enum ColumnWidthValues {
   AutoMin = 80,
 }
 
+const isBold = ({ fontWeight }: CellStyling) => fontWeight !== "normal";
```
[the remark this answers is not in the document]
cbt1: We will support that, but we don't currently. It can only be `600`, `normal` or `undefined` at the momemt

test/rendering/render.mjs-snapshots/scenario-1-chromium-linux.png
cbt1, line 1: Had to update the baseline images because bold is now used when measuring the dimension values. Which means it takes up more space in the scenarios where the dimension values are wider then the header.

src/pivot-table/hooks/use-column-width.ts
a-m-dev: btw, this one should wait for `useMeasureText` update from shared lib, right?

cbt1: No. I don't think there is a reason to wait. The currently logic works just fine. But we probably want to add the option to set font weight through the theme. Then this logic will not work as the user can set other values than `600/normal/undefined`.

I said probably here as well since the interaction between the `font-weight` value from the theme and the styling panels "bold" button is going to be tricky at best. It would also include dealing with custom `font-style` and `text-decoration`.

PR #412, opened 2023-10-16
fix: argb to rgba

cbt1:
The alpha value of `ARGB` goes from 0 to 255, while the alpha value of `RGBA` goes from 0 to 1. This fix normlizes the `ARGB` alpha value and also extract it correctly from the color string. Previously it was set as the `R` value.

EDIT: The actual fix is now made in the utils repo


test/rendering/__fixtures__/scenario-color-by-expression.fix.js
```diff
@@ -302,7 +302,7 @@ export default () => ({
                       qAttrExps: {
                         qValues: [
                           {
-                            qText: "ARGB(6,118,138,70)",
+                            qText: "ARGB(255,118,138,70)",
```
cbt1, line 305: The low alpha preduced a background that was almost white, but the `isDarkColor` logic does not take alpha into account so it thinks the color is dark. Which resulted in a "white" background color with a white text color.

a-m-dev (approved):

PR #414, opened 2023-10-16
fix: primary text color

cbt1:
Changes so that the default text color is resolved from the sprout color palette instead of being hard-coded in sn-pivot-table repo.

Currently the colors are the same and I would guess not really expected to change any time soon.

### DIFF
--- a/src/pivot-table/components/shared-styles.ts
+++ b/src/pivot-table/components/shared-styles.ts
@@ -2,6 +2,7 @@ import type React from "react";
 import type { ShowLastBorder } from "../../types/types";
 import { LINE_HEIGHT_COEFFICIENT } from "../constants";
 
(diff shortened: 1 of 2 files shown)


a-m-dev (commented):
couldn't you just replace the value in `Colors.PrimaryText` in shared styled?

cbt1:
> couldn't you just replace the value in `Colors.PrimaryText` in shared styled?

I think we want to move away from the `Colors` enum and consume colors from sprout instead, either via utils repo or direct from sprout

a-m-dev (approved):

a-m-dev:
then IMO that would be nice to have a `// TODO` on top of the `Colors` enum, hinting about we should not rely on it when try to use it

PR #416, opened 2023-10-16
refactor: use headers data for column width

enell:
To prepare for horizontal headers make column width
use headers data instead of visibleLeftDimensionInfo

Refs: VIZ-2530

### DIFF
--- a/src/pivot-table/components/grids/HeaderGrid.tsx
+++ b/src/pivot-table/components/grids/HeaderGrid.tsx
@@ -40,6 +40,8 @@ const HeaderGrid = ({
 
   const hasMultipleRows = headersData.size.y > 1;
 
+  const lastRow = headersData.data.at(-1) ?? [];
+
   return (
     <div
       style={{
@@ -51,8 +53,8 @@ const HeaderGrid = ({
       }}
     >
       {hasMultipleRows && <EmptyHeaderCell columnWidths={columnWidths} />}
-      {headersData.data.map((col, colIndex) => {
-        const cell = col[col.length - 1] as HeaderCell;
+      {lastRow.map((col, colIndex) => {
+        const cell = col as HeaderCell;
         const iconsVisibilityStatus = getHeaderCellsIconsVisibilityStatus(colIndex, cell.isLocked, cell.label);
 
         return (
--- a/src/pivot-table/data/extract-headers.ts
+++ b/src/pivot-table/data/extract-headers.ts
@@ -5,17 +5,20 @@ import getKey from "../components/helpers/get-key";
 
 const extractHeaders = (
   hyperCube: ExtendedHyperCube,
-  rowCount: number,
+  visibleTopDimensionInfo: VisibleDimensionInfo[],
   visibleLeftDimensionInfo: VisibleDimensionInfo[],
 ): (null | HeaderCell)[][] => {
-  const matrix: (null | HeaderCell)[][] = Array(visibleLeftDimensionInfo.length)
+  // rowCount cannot be 0, as it will cause an issue when there is no top data but there is left data
+  const rowCount = Math.max(1, visibleTopDimensionInfo.length);
+  const matrix: (null | HeaderCell)[][] = Array(rowCount)
     .fill(null)
-    .map(() => Array.from({ length: rowCount }, () => null));
+    .map(() => Array.from({ length: visibleLeftDimensionInfo.length }, () => null));
 
   visibleLeftDimensionInfo.forEach((qDimensionInfo, colIdx) => {
+    const id = getKey(qDimensionInfo);
     if (qDimensionInfo === PSEUDO_DIMENSION_INDEX) {
-      matrix[colIdx][rowCount - 1] = {
-        id: "PSEUDO-DIM",
+      matrix[rowCount - 1][colIdx] = {
+        id,
         colIdx: -1,
         label: "",
         sortDirection: "A",
@@ -26,8 +29,7 @@ const extractHeaders = (
         headTextAlign: "left",
       };
     } else {
-      const id: string = getKey(qDimensionInfo);
-      matrix[colIdx][rowCount - 1] = {
+      matrix[rowCount - 1][colIdx] = {
         id,
         colIdx,
         label: qDimensionInfo.qFallbackTitle,
@@ -38,6 +40,8 @@ const extractHeaders = (
         fieldId: qDimensionInfo.qGroupFieldDefs[qDimensionInfo.qGroupPos],
         isActivelySorted: colIdx === (hyperCube.activelySortedColumn?.colIdx ?? 0),
         isLocked: qDimensionInfo.qLocked ?? false,
+        columnWidth: qDimensionInfo.columnWidth,
+        qApprMaxGlyphCount: qDimensionInfo.qApprMaxGlyphCount,
         isDim: true,
         headTextAlign: "left",
       };
--- a/src/pivot-table/hooks/use-column-width.ts
+++ b/src/pivot-table/hooks/use-column-width.ts
@@ -1,8 +1,15 @@
 import { useMeasureText } from "@qlik/nebula-table-utils/lib/hooks";
 import { useCallback, useMemo } from "react";
-import { PSEUDO_DIMENSION_INDEX } from "../../constants";
+import { PSEUDO_DIMENSION_INDEX, PSEUDO_DIMENSION_KEY } from "../../constants";
 import { ColumnWidthType, type ColumnWidth } from "../../types/QIX";
-import type { CellStyling, LayoutService, Rect, VisibleDimensionInfo } from "../../types/types";
+import type {
+  CellStyling,
+  HeaderCell,
+  HeadersData,
+  LayoutService,
+  Rect,
+  VisibleDimensionInfo,
+} from "../../types/types";
 import { CELL_PADDING } from "../components/shared-styles";
 import { GRID_BORDER, HEADER_ICON_SIZE } from "../constants";
 import { useStyleContext } from "../contexts/StyleProvider";
@@ -56,7 +63,7 @@ const getPercentageValue = (percentage: number | undefined) =>
 export default function useColumnWidth(
   layoutService: LayoutService,
   rect: Rect,
-  visibleLeftDimensionInfo: VisibleDimensionInfo[],
+  headersData: HeadersData,
   visibleTopDimensionInfo: VisibleDimensionInfo[],
 ): ColumnWidthHook {
   const {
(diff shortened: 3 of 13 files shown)


src/pivot-table/data/extract-headers.ts
```diff
@@ -5,17 +5,18 @@ import getKey from "../components/helpers/get-key";
...
     .fill(null)
-    .map(() => Array.from({ length: rowCount }, () => null));
+    .map(() => Array.from({ length: visibleLeftDimensionInfo.length }, () => null));
 
   visibleLeftDimensionInfo.forEach((qDimensionInfo, colIdx) => {
     if (qDimensionInfo === PSEUDO_DIMENSION_INDEX) {
-      matrix[colIdx][rowCount - 1] = {
-        id: "PSEUDO-DIM",
+      matrix[rowCount - 1][colIdx] = {
+        id: `${PSEUDO_DIMENSION_INDEX}`,
```
cbt1, line 19: Consider setting `isPseudoDimension: true` property instead of trying to check against the ID if the cell is a pseudo dimension. Then there should be no need to change the `id` or do the change in `getKey`

enell: isPseudoDimension is part of the Cell type. Not the HeaderCell type

This change was to use the same id for pseudo cell instead of introducing a new

src/pivot-table/components/grids/HeaderGrid.tsx
```diff
@@ -40,6 +40,8 @@ const HeaderGrid = ({
 
   const hasMultipleRows = headersData.size.y > 1;
 
+  const lastRow = () => headersData.data.at(-1) ?? [];
```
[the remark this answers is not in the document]
enell: Fixed

src/pivot-table/components/cells/DimensionTitleCell.tsx
```diff
@@ -64,7 +64,7 @@ const DimensionTitleCell = ({
   const { shouldShowLockIcon, shouldShowMenuIcon } = iconsVisibilityStatus;
   const { open, setOpen, handleOpenMenu } = useHeadCellDim({ interactions });
 
-  const isDim = cell.id !== "PSEUDO-DIM";
+  const isDim = cell.id !== "-1";
```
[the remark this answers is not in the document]
enell: I introduced a constant for the pseudo dimension key instead. It is not the same as the index.

src/pivot-table/data/extract-headers.ts
enell: Changing type from HeaderCell to Cell is better handled by another PR

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -110,14 +118,14 @@ export default function useColumnWidth(
...
         const iconWidth = !isFullyExpanded && index < qNoOfLeftDims - 1 ? EXPAND_ICON_WIDTH : 0;
         const fitToContentWidth = Math.max(
-          measureTextForHeader(qFallbackTitle) + TOTAL_CELL_PADDING,
-          estimateWidthForDimensionValue(qApprMaxGlyphCount) + iconWidth,
+          measureTextForHeader(title) + TOTAL_CELL_PADDING,
+          estimateWidthForDimensionValue(qApprMaxGlyphCount as number) + iconWidth,
         );
 
-        width = getColumnWidth(columnWidth, fitToContentWidth);
+        width = getColumnWidth(columnWidth as ColumnWidth, fitToContentWidth);
```
cbt1, line 128: On main `columnWidth` can be undefined. Which should be handled by `getColumnWidth`. So there should be no need to cast it here

src/pivot-table/data/extract-headers.ts
```diff
@@ -24,18 +26,19 @@ const extractHeaders = (
...
+      matrix[rowCount - 1][colIdx] = {
         id,
         colIdx,
         title: qDimensionInfo.qFallbackTitle,
         qReverseSort: qDimensionInfo.qReverseSort,
         sortDirection:
           qDimensionInfo.qSortIndicator && qDimensionInfo.qSortIndicator !== "N" ? qDimensionInfo.qSortIndicator : "A",
         qLibraryId: qDimensionInfo.qLibraryId,
-        fieldId: qDimensionInfo.qGroupFieldDefs[qDimensionInfo.qGroupPos],
+        fieldId: qDimensionInfo.qGroupFieldDefs ? qDimensionInfo.qGroupFieldDefs[qDimensionInfo.qGroupPos] : "",
```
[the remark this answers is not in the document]
enell: qGroupFieldDefs isn't optional but based on what you said before about qLocked I can't trust the EngineAPI type defenition. So if qLocked can be undefined any property can be undefined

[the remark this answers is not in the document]
enell: Or do you mean that I should change the default value to undefined?

src/pivot-table/hooks/use-column-width.ts
enell: fixed

src/pivot-table/data/extract-headers.ts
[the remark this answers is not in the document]
enell: I can't change it. It ripples back to HeaderData type from table utils repo and I can't change that in this pr

src/pivot-table/data/extract-headers.ts
```diff
@@ -5,17 +5,20 @@ import getKey from "../components/helpers/get-key";
...
+  visibleTopDimensionInfo: VisibleDimensionInfo[],
   visibleLeftDimensionInfo: VisibleDimensionInfo[],
 ): (null | HeaderCell)[][] => {
-  const matrix: (null | HeaderCell)[][] = Array(visibleLeftDimensionInfo.length)
+  // rowCount cannot be 0, as it will cause an issue when there is no top data but there is left data
+  const rowCount = Math.max(1, visibleTopDimensionInfo.length);
+  const matrix: (null | HeaderCell)[][] = Array(rowCount)
     .fill(null)
-    .map(() => Array.from({ length: rowCount }, () => null));
+    .map(() => Array.from({ length: visibleLeftDimensionInfo.length }, () => null));
```
a-m-dev, line 15: I think you are shifting the matrix here, 
the original matrix is `leftDimInfo.length` * `topDimRowCounts`
but with your changes it's now the `topDimRowCounts` * `leftDimInfo.length`

what i noticed is this file path here: 
1) in use-data.ts file we call [`createHeadersData()`](https://github.com/qlik-oss/sn-pivot-table/blob/df0c53a317fe4fa0fda7e18165692ed82edc1094/src/pivot-table/hooks/use-data.ts#L148) and then pass topDimensionData.rowCount as rowCount 
2) within [`createHeadersData()`](https://github.com/qlik-oss/sn-pivot-table/blob/df0c53a317fe4fa0fda7e18165692ed82edc1094/src/pivot-table/data/headers-data.ts#L11), we forward that as second argument to `extractHeaders(..., rowCount, ...)`
3) and then, in `extractHeaders()` we use that rowCount as [width of matrix at here](https://github.com/qlik-oss/sn-pivot-table/blob/df0c53a317fe4fa0fda7e18165692ed82edc1094/src/pivot-table/data/extract-headers.ts#L13C36-L13C36) and then `visibleLeftDimensionInfo.length` as height of matrix

what happens in here is that you are using `visibleTopDimensionInfo.length` as height and `visibleLeftDimensionInfo.length` as width of the resulted matrix

is that causing any issues? (or may be my brain is broken 🥴 )

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -110,11 +118,11 @@ export default function useColumnWidth(
...
         );
       } else {
-        const { qFallbackTitle, qApprMaxGlyphCount, columnWidth } = qDimensionInfo;
+        const { title, qApprMaxGlyphCount, columnWidth } = cell;
         const iconWidth = !isFullyExpanded && index < qNoOfLeftDims - 1 ? EXPAND_ICON_WIDTH : 0;
         const fitToContentWidth = Math.max(
-          measureTextForHeader(qFallbackTitle) + TOTAL_CELL_PADDING,
-          estimateWidthForDimensionValue(qApprMaxGlyphCount) + iconWidth,
+          measureTextForHeader(title) + TOTAL_CELL_PADDING,
+          estimateWidthForDimensionValue(qApprMaxGlyphCount as number) + iconWidth,
```
a-m-dev, line 125: is there a need to case `as number` here? 
since the type is number and seems like we should always have it's value according [to here](https://github.com/qlik-oss/sn-pivot-table/blob/1a37c577fdcded0081f76ce08de681726bd91d12/src/pivot-table/data/extract-headers.ts#L42) 

I noticed in types [that is optional](https://github.com/qlik-oss/sn-pivot-table/blob/1a37c577fdcded0081f76ce08de681726bd91d12/src/types/types.ts#L28) but is it because of that you need to cast it here?

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -98,10 +105,11 @@ export default function useColumnWidth(
 
     let sumOfWidths = 0;
 
-    const columnWidths = visibleLeftDimensionInfo.map((qDimensionInfo, index) => {
+    const lastRow = headersData.data[headersData.size.y - 1] as HeaderCell[];
+    const columnWidths = lastRow.map((cell, index) => {
```
a-m-dev, line 109: what if we have this case as header? then the title value on last yellow cell on headers data will be an empty string and the calculations about `fitToContentWidth` will be wrong i guess? 
(I think that might be the reason why we only rely on leftGrid data/visibleLeftDimensionInfo instead of only headers data) or maybe i'm missing something 

Screenshot 2023-10-18 at 13 25 20

src/pivot-table/data/extract-headers.ts
enell: It correct observartion that I'm shifting the matrix. That is to make it easier when I use it later in the HeaderGrid. I want to iterate each row and for each cell in the row I want to place an element in the grid. I can shift the order of the props so that it makes sense in all function calls

src/pivot-table/hooks/use-column-width.ts
enell: Yes. I'm casting it becaus it is optional. I know that it is defined when the cell isn't a pseudo dimension which I already checked above

src/pivot-table/data/extract-headers.ts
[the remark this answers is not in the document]
enell: That is a very convinsing answer. I will change qGroupFieldDefs. But also, I will see if qLocked should be updated to optional

src/pivot-table/data/extract-headers.ts
a-m-dev: as we discussed, the scope of this only falls into header grid and since you are having an eye on it's side effect then it should be fine!

src/pivot-table/hooks/use-column-width.ts
a-m-dev: as of our discussion, seems like your changes working fine with headers on pseudo dims (in a scenario where there is no title to calculate the width of cell)

a-m-dev (approved):
LGTM 🚀

PR #417, opened 2023-10-16
fix: overlay scrollbars on data

a-m-dev:
Including in this PR: 
- [x] fix for overlay vertical scrollbar on data
- [x] fix for overlay horizontal scrollbar on data
- [x] (lots of) testing, both manual, unit and e2e tests to make sure all good

**what is missing?**
Testing on windows machine ☠️

---

Below are a few important edge case scenarios for this PR, (if you try to test make sure to check these special cases): 

**1) should not preserve horizontal scrollbar space when it disappears during resize:** 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/1bba7967-f2b6-4640-a487-35ba6658cb29

**2) only either of grids (left || data grid) having scrollbar:** 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/3bfeff20-8053-42a8-a114-974373057756

**3) horizontal scrollbar in the last page:** 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/19d63240-9e05-4072-a0b1-2c6aa6d5575c

**4) having a very loooong horizontal pivot table:** 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/6b590771-4c93-4bee-8cc6-bf0a472a8619

**5) vertical/horizontal scrollbar might be excluded from calculations:** 
basically if you check all above recordings, always we have correct space preservation in our calculation in y and x axis -> NEVER SCROLLBAR OVERLAYS ON DATA -> it always has it's own space

**6) the very corner cell might always be visible:** 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/6f715831-a6a6-49f2-86ec-6847fb5e3705

**7) expand col(s):** 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/a2ed2b98-6a6e-4f7b-8b39-67383559f75b

**8) <YOUR_SPECIAL_CASE_HERE>**
...

**Please Note that:** 
In all above scenarios **scrollbar might not overlay on data and block it at any cost** because that is the goal of this PR


src/pivot-table/hooks/use-column-width.ts
```diff
@@ -306,6 +316,15 @@ export default function useColumnWidth(
 
   const showLastRightBorder = useMemo(() => totalWidth < rect.width, [totalWidth, rect.width]);
 
+  // Horizontal scrollbar height control based on columns (full) visibility
+  useEffect(() => {
+    const allLeftGridColumnsVisible = leftGridWidthInfo.leftGridWidth === leftGridWidthInfo.leftGridFullWidth;
+    const allDataGridColumnsVisible = rightGridWidth === rightGridFullWidth;
+
+    if (allLeftGridColumnsVisible && allDataGridColumnsVisible) horizontalScrollbarHeightSetter(true);
```
cbt1, line 324: curly braches?

a-m-dev: it felt unnecessary since it's just only one line, but will fix it

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -306,6 +316,15 @@ export default function useColumnWidth(
...
+    else horizontalScrollbarHeightSetter();
+  }, [leftGridWidthInfo, rightGridWidth, rightGridFullWidth, horizontalScrollbarHeightSetter]);
```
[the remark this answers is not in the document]
a-m-dev: one possible solution could be something like this:
```tsx
  // Horizontal scrollbar height control based on columns (full) visibility
  const allLeftGridColumnsVisible = leftGridWidthInfo.leftGridWidth === leftGridWidthInfo.leftGridFullWidth;
  const allDataGridColumnsVisible = rightGridWidth === rightGridFullWidth;
  
  useEffect(() => {
    horizontalScrollbarHeightSetter(allLeftGridColumnsVisible && allDataGridColumnsVisible);
  }, [allLeftGridColumnsVisible, allDataGridColumnsVisible]);
```

will investigate on that a bit more

src/pivot-table/hooks/use-column-width.ts
cbt1: I understand, but for consistency I think we should use it

src/pivot-table/components/helpers/get-scrollable-areas-dimensions.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  return {
+    ROOT_WRAPPER: {
+      scrollable: tableRect,
+      fullSize: {
+        width: totalWidth - modifiedVerticalScrollbarWidth,
+        height: containerHeight + modifiedHorizontalScrollbarHeight,
+      },
+      sticky: {
+        width: totalWidth - modifiedVerticalScrollbarWidth,
```
cbt1, line 71: Would does the sticky container have the full width? Should it not be `tableRect.width - modifiedVerticalScrollbarWidth`?

src/pivot-table/components/helpers/get-scrollable-areas-dimensions.ts
```diff
@@ -0,0 +1,126 @@
...
+  const modifiedVerticalScrollbarWidth = verticalScrollbarWidth * (allRowsVisible ? 0 : 1);
+  const modifiedHorizontalScrollbarHeight = horizontalScrollbarHeight * (allRowsVisible ? 0 : 1);
+
+  const childWrappersScrollableContainerHeight = allRowsVisible
+    ? topGridHeight + dataGridHeight + GRID_BORDER + horizontalScrollbarHeight
+    : tableRect.height;
+  const childWrappersStickyContainerHeight = tableRect.height - modifiedHorizontalScrollbarHeight;
+
+  return {
+    ROOT_WRAPPER: {
```
cbt1, line 64: All caps tend to indicate that it's a static value/constant. But this is clearly not, so consider renaming it to `rootWrapper`

src/pivot-table/components/helpers/__tests__/get-scrollable-areas-dimensions.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      dataGridHeight,
+      allRowsVisible,
+      totalWidth,
+      leftGridWidth,
+      rightGridWidth,
+      verticalScrollbarWidth,
+      horizontalScrollbarHeight,
+    });
+
+  test("check result object structure to be correct", () => {
```
cbt1, line 48: This seems redundent since you other tests are already checking all possible properties

src/pivot-table/hooks/use-column-width.ts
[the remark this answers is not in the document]
cbt1: `horizontalScrollbarHeightSetter` will need to be in the deps array. Wrap `horizontalScrollbarHeightSetter` in a callback should be sufficient?

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -142,6 +146,7 @@ export default function useColumnWidth(
     return {
       leftGridWidth: Math.min(rect.width * LEFT_GRID_MAX_WIDTH_RATIO, sumOfWidths),
       leftGridColumnWidths: columnWidths,
+      leftGridFullWidth: columnWidths.reduce((acc, curr) => acc + curr, 0),
```
cbt1, line 149: This should always be the same as `sumOfWidths?`

cbt1: In which case you might as well expose it as `allLeftGridColumnsVisible` here instead

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -282,11 +287,16 @@ export default function useColumnWidth(
...
-        : averageLeafWidth,
-    [topGridLeavesIsPseudo, leafWidths, layoutService, averageLeafWidth],
+    (index?: number) => {
+      const scrollbarWidthSharePerColumn = !allRowsVisible
+        ? parseFloat((verticalScrollbarWidth / layoutService.size.x).toFixed(12))
+        : 0;
+
+      return topGridLeavesIsPseudo && index !== undefined
+        ? leafWidths[layoutService.getMeasureInfoIndexFromCellIndex(index)] - scrollbarWidthSharePerColumn
+        : averageLeafWidth - scrollbarWidthSharePerColumn;
```
haxxmaxx, line 297: How come we remove the scrollbar width at this point? Can't you do it when you calculate leafWidths? By just subtracting the width from `rightGridAvailableWidth`?

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -70,6 +71,9 @@ export default function useColumnWidth(
   rect: Rect,
   headersData: HeadersData,
   visibleTopDimensionInfo: VisibleDimensionInfo[],
+  allRowsVisible: boolean,
+  verticalScrollbarWidth: number,
```
haxxmaxx, line 75: Not sure if this is a timing thing, isn't this true?` allRowsVisible` -> `verticalScrollbarWidth === 0` or `!allRowsVisible` -> `verticalScrollbarWidth > 0`. So you only need one of them?

src/pivot-table/components/helpers/get-scrollable-areas-dimensions.ts
a-m-dev: Absolutely Correct! 
I haven't noticed it during my changes just because the sticky wrapper always remains sticky in the screen.

src/pivot-table/hooks/use-column-width.ts
a-m-dev: Yes, they are same, but for naming i think it should stay like as it is (`leftGridFullWidth`) since we have it's counterpart for right grid as well at here: https://github.com/qlik-oss/sn-pivot-table/blob/472e5ae2b958312641ecad224f002c1a88299fe2/src/pivot-table/hooks/use-column-width.ts#L293

src/pivot-table/hooks/use-column-width.ts
a-m-dev: because getRightGridColumnWidth() is the unified way we calculate columns width one by one -> by passing it to the VT related components. 

The point is we dont have a single value indicating the entire width of **all** columns at once, (we may calculate it at some point but it's not been used in VT components to define the width), every single column width being calculated one by one and that is why I chosed to go this way.

Plus, `rightGridAvailableWidth` is the viewable area in right grid, not the (full) width we use for columns, which is the final guy who decides on (sum) width of all columns

src/pivot-table/hooks/use-scroll.ts
```diff
@@ -77,7 +125,7 @@ const useScroll = ({ layoutService, pageInfo, mockedRefs }: Props) => {
     if (!(evt.target instanceof HTMLDivElement)) return;
 
     if (leftGridRef.current) {
-      leftGridRef.current.forEach((list) => list.scrollTo(evt.currentTarget.scrollTop));
+      leftGridRef.current.filter(Boolean).forEach((list) => list.scrollTo(evt.currentTarget.scrollTop));
```
cbt1, line 128: Why the `filter` here?

a-m-dev: because when we expand a node we get this list also expanded (from `list: [{col#01}]` to `list: [{col#01}, {col#02}, ...]`)

when we collapse the nodes -> the list is not shrinking down to `list: [{col#01}]`, but it stays like `list: [{col#01}, null, ...]`). So that null values causes this to throw error complaining about `null.scrollTo(...)` does not exists and this filter boolean just skips those null values

cbt1: Has that changed? Has the`topGridRef` not the same issue?

a-m-dev: top grid is working fine, but that only was in case of leftGrid

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/80e3d7f3-93b5-4ec1-a9ef-1aad0a7c8ec2

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: So i gave my idea a try. Didn't get the calculations correct, so I think we can keep the `- scrollbarWidthSharePerColumn` here

src/pivot-table/hooks/use-column-width.ts
a-m-dev: fixed

haxxmaxx (commented):
The column width parts looks good to me, I'll leave the rest of the review for Christofer

cbt1 (approved):

PR #418, opened 2023-10-17
fix: column width error

cbt1:
Fixes an issue where an error would be thrown when re-sizing the bottom top dimension, if the left side had a pseudo dimension.

Also fixes the type definition. As `columnWidth` does not exist on sn-pivot-tables created in older versions.

### DIFF
--- a/src/pivot-table/hooks/use-data-model.ts
+++ b/src/pivot-table/hooks/use-data-model.ts
@@ -84,6 +84,8 @@ export default function useDataModel({
 
   const applyColumnWidth = useCallback<ApplyColumnWidth>(
     (newColumnWidth: ColumnWidth, { isPseudoDimension, isAncestorPseudoDimension, isLeftColumn, x, y }: Cell) => {
+      const { qNoOfLeftDims, qMeasureInfo, qDimensionInfo } = layoutService.layout.qHyperCube;
+
       let index: number;
       if (isPseudoDimension) {
         // TODO: what todo with the left pseudo dim? set all measures?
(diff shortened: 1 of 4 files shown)


src/pivot-table/hooks/use-data-model.ts
```diff
@@ -84,19 +84,21 @@ export default function useDataModel({
...
         index = isLeftColumn
           ? x - Number(isAncestorPseudoDimension)
-          : y - Number(isAncestorPseudoDimension) + layoutService.layout.qHyperCube.qNoOfLeftDims;
+          : y - Number(isAncestorPseudoDimension) + qNoOfLeftDims - Number(layoutService.hasPseudoDimOnLeft);
       }
 
       const qPath = `${Q_PATH}/${isPseudoDimension ? "qMeasures" : "qDimensions"}/${index}/qDef/columnWidth`;
-      const oldColumnWidth =
-        layoutService.layout.qHyperCube[isPseudoDimension ? "qMeasureInfo" : "qDimensionInfo"][index].columnWidth;
+      const oldColumnWidth = isPseudoDimension ? qMeasureInfo[index].columnWidth : qDimensionInfo[index].columnWidth;
```
haxxmaxx, line 100: this is way nicer

haxxmaxx (approved):
So this only happens if only the pseudo dim is on the left? not if you have another dimension on the left as well?

cbt1:
> So this only happens if only the pseudo dim is on the left? not if you have another dimension on the left as well?

Did some more testing and it always crashes when you have pseudo dim on the left side. With or without more dimensions there

PR #419, opened 2023-10-18
fix: center align top grid cell content

cbt1:
According to the design spec. The top grid cell content should be center aligned along the x-axis. This PR fixes that.

I have not been able to keep the content centered while scrolling. As seen in the clip below. I could not come up with a pure CSS solution for that. Otherwise the content would need to be re-rendered via React while scrolling which would be less then ideal for the performance.

DEMO:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/ba192515-a7f4-48e8-85be-cf159b2f64c4


src/pivot-table/components/cells/utils/get-dimension-cell-style.ts
```diff
@@ -108,13 +109,15 @@ export const getContainerStyle = ({
     ...resolvedLockedSelectionStyle,
     display: "flex",
     zIndex,
+    justifyContent: isLeftColumn ? undefined : "center",
```
a-m-dev, line 112: you can do the same thing here: 
```suggestion
    ...(!isLeftColumn && { justifyContent: "center" })
```

src/pivot-table/components/cells/utils/get-dimension-cell-style.ts
```diff
@@ -108,13 +109,15 @@ export const getContainerStyle = ({
...
   };
 };
 
 export const getInnerContainerStyle = (isLeftColumn: boolean) => ({
   ...cellStyle,
   ...stickyCell,
   alignSelf: isLeftColumn ? "flex-start" : "center",
+  right: isLeftColumn ? undefined : CELL_PADDING,
```
a-m-dev, line 120:
```suggestion
    ...(!isLeftColumn && { right: CELL_PADDING })
```

a-m-dev (commented):
when texts are ellipses, seems like the center alignment is not applied: 
Screenshot 2023-10-19 at 09 52 04

cbt1:
> when texts are ellipses, seems like the center alignment is not applied: 


It is applied but the ellipses logic is broken. The div with the text is taking up the whole width of the cell but the text is elliped at the wrong position.

It is the same on main.

a-m-dev (approved):

a-m-dev:
Probably we might take a look at that issue in another PR. 
(the wrapper div for ellipsis should be centered as well since seems like there are some space to it's right)

PR #421, opened 2023-10-19
fix: include menu icons in column width

cbt1:
Includes the menu icons width for FitToContent and Auto column widths.

Before
Skärmavbild 2023-10-18 kl  14 15 38

After
Skärmavbild 2023-10-18 kl  14 15 56


a-m-dev (approved):
LGTM, 
I wish there was a scenario for locked field state as well just to make sure that we are not ignoring the important of locked icon over menu icon in limited column width cases

PR #422, opened 2023-10-19
fix: text ellipsing

cbt1:
Fixes an issue where the text was always ellipsed at the end of a word instead of at any position in the word.


src/pivot-table/components/shared-styles.ts
```diff
@@ -48,6 +48,9 @@ export const getLineClampStyle = (clampCount: number): React.CSSProperties => ({
   WebkitLineClamp: clampCount,
   WebkitBoxOrient: "vertical",
   lineBreak: "auto",
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+  wordBreak: "break-all",
```
cbt1, line 53: This is what actually fixes the issue. Every thing else is just refactoring

haxxmaxx (approved):

PR #427, opened 2023-10-23
fix: handle failed selections

cbt1:
This is a refactoring that turned into a fix when I found 2 bugs in the code.

Bugs fixed:
- Do not update internal selection state when the `select` method to Engine fails
- An error was thrown when you clicked on the "expand/collapse" icon while in selection mode.

Refactors:
- Simplify the method from the selection model so that a `Cell` can be passed as paremter. This is to reduce the number of expressions in the `DimensionCell` component.
- Use `Set` instead of an `Array` to keep track of selected cells. This should improve the performance when scrolling while being in selection mode, as each render of `DimensionCell` no longer need to do multiple O(n) lookups. It can instead enjoy a breezing O(1) lookup.


src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -48,68 +86,68 @@ export default function useSelectionsModel(
...
     },
-    [selected],
+    [selectedPivotCells, selectedRowOrColumn],
   );
 
   const select = useCallback(
-    (qType: EngineAPI.NxSelectionCellType, qRow: number, qCol: number) => async (evt: React.MouseEvent) => {
-      if ((evt.target as HTMLElement)?.className.includes("sn-pivot-table-column-adjuster")) {
+    (cell: Cell) => async (evt: React.MouseEvent) => {
+      if ((evt.target as HTMLElement | SVGElement)?.getAttribute("class") === "sn-pivot-table-column-adjuster") {
```
cbt1, line 109: When clicking on the expand/collapse icon, you get an SVG element. Which does not behave the same when `className` is called (https://developer.mozilla.org/en-US/docs/Web/API/Element/className#notes)

src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -48,68 +86,68 @@ export default function useSelectionsModel(
...
+        selectedPivotCells,
+        selectedRowOrColumn,
+      );
+
+      try {
+        await selections.select({
+          method: "selectPivotCells",
+          params: [
+            Q_PATH,
+            Array.from(nextSelectedPivotCells.values()).map((c) => ({
```
a-m-dev, line 132:
```suggestion
            Array.from(nextSelectedPivotCells).map((c) => ({
```
no `.values()` needed

a-m-dev: also a better suggestion, I think you can even do this:
```suggestion
            Array.from(nextSelectedPivotCells, (c) => ({ 
                qType: c.selectionCellType,
                qRow: c.y,
                qCol: c.x,
             })
            )
```

because `Array.from(...)` accepts a 2nd arg of a transformer

src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -48,68 +86,68 @@ export default function useSelectionsModel(
...
+
+      if (cell.selectionCellType !== selectedRowOrColumn.selectionCellType) {
+        return true;
+      }
+
+      if (cell.selectionCellType === NxSelectionCellType.NX_CELL_LEFT) {
+        return selectedRowOrColumn.coord !== cell.x;
+      }
+
+      return selectedRowOrColumn.coord !== cell.y;
```
a-m-dev, line 102: Maybe this is like a qlik specific question, but out of curiosity + i don't understand this part that why not matching coords of a cell would means that cell is locked or not? 

I see (kinda) same logic in the previous version of `isLocked` implementation as well

src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -48,68 +86,68 @@ export default function useSelectionsModel(
...
+            Array.from(nextSelectedPivotCells.values()).map((c) => ({
+              qType: c.selectionCellType,
+              qRow: c.y,
+              qCol: c.x,
+            })),
+          ],
+        });
+
+        setSelectedPivotCells(nextSelectedPivotCells);
+        setSelectedRowOrColumn(nextSelectedRowOrColumn);
```
a-m-dev, line 141: ~~doesn't these two line replace the same value to states?
because not transformation happens to values of: `nextSelectedPivotCells` and `nextSelectedRowOrColumn` since they been retrieved from `getNextState()` on line 121~~

isn't it better to keep state update and data transformation at the same place? (just to not fall into 👆 trap while reading the code)

src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -21,17 +19,57 @@ export interface SelectedPivotCell {
...
   selections: ExtendedSelections,
   updatePageInfo: (args: Partial<PageInfo>) => void,
 ): SelectionModel {
   const isActive = selections.isActive();
-  const [selected, setSelected] = useState<SelectedPivotCell[]>([]);
+  const [selectedPivotCells, setSelectedPivotCells] = useState<Set<Cell>>(new Set());
+  const [selectedRowOrColumn, setSelectedRowOrColumn] = useState<SelectedRowOrColumn>({
+    selectionCellType: null,
+    coord: -1,
+  });
```
a-m-dev, line 64: this value will only going to be point to the last cell that processed and keep itself updated by it's info, but within the isLocked function (which could receive **any** cell) we rely on (last processed) value to compute the locked state (line 89 - 102). 

(or maybe) am i miss-interpreting the changes?

src/pivot-table/hooks/use-selections-model.ts
cbt1: Ah yes, good find. I initially used a `Map` instead of `Set`. But did not think of updating these

src/pivot-table/hooks/use-selections-model.ts
cbt1: You can only do selections in one dimension field at the time. So if there are ongoing selections in a field. All other fields are locked.

But since field information is not available here. We use the `selectionCellType` and `coord` to figure out which field have ongoing selections.

src/pivot-table/hooks/use-selections-model.ts
cbt1: We first need to transform the data, send it to Engine and if all goes well we can update the state. So I don't think there is a way to keep "state update" and "data transformation" in the same place?

src/pivot-table/hooks/use-selections-model.ts
cbt1: I think the answer in one of the other comments should answer this as well?

src/pivot-table/hooks/use-selections-model.ts
a-m-dev: can we have a comment about it, or it's to simple to have it like this (because i get confused tbh for a sec 😰)

a-m-dev (approved):
LGTM 🚢

src/pivot-table/components/cells/DimensionCell.tsx
```diff
@@ -129,7 +127,7 @@ const DimensionCell = ({
     expressionBackground: cell.expressionColor.background,
     zIndex: layoutService.size.x - colIndex,
   });
-  const onClickHandler = isNonSelectableCell ? undefined : select(selectionCellType, cell.y, colIndex);
+  const onClickHandler = isNonSelectableCell ? undefined : select(cell);
```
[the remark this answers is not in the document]
cbt1: Will revert this change as you get Reach warnings from it:
> Warning: Expected `onClick` listener to be a function, instead got `false`.

PR #429, opened 2023-10-23
fix: plus minus icon size

a-m-dev:
Plus/minus icon shrinking in the size when text requires bigger space. This PR fixes it.

Before: 
Screenshot 2023-10-23 at 14 50 43

After:
Screenshot 2023-10-23 at 14 50 06

### DIFF
--- a/src/pivot-table/components/icons/styles.ts
+++ b/src/pivot-table/components/icons/styles.ts
@@ -1,0 +1,7 @@
+import { HEADER_ICON_SIZE } from "../../constants";
+import { baseFlex } from "../shared-styles";
+
+export const headerPlusMinusIconWrapperStyle = {
+  width: HEADER_ICON_SIZE,
+  ...baseFlex,
+};
--- a/src/pivot-table/components/icons/styles.ts
+++ b/src/pivot-table/components/icons/styles.ts
@@ -1,7 +1,6 @@
-import { HEADER_ICON_SIZE } from "../../constants";
 import { baseFlex } from "../shared-styles";
 
 export const headerPlusMinusIconWrapperStyle = {
-  width: HEADER_ICON_SIZE,
+  width: 13,
   ...baseFlex,
 };
--- a/src/pivot-table/components/icons/styles.ts
+++ b/src/pivot-table/components/icons/styles.ts
@@ -1,6 +1,7 @@
+import { PLUS_MINUS_ICON_SIZE } from "../../constants";
 import { baseFlex } from "../shared-styles";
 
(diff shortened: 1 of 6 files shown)


src/pivot-table/components/icons/styles.ts
```diff
@@ -0,0 +1,7 @@
+import { HEADER_ICON_SIZE } from "../../constants";
+import { baseFlex } from "../shared-styles";
+
+export const headerPlusMinusIconWrapperStyle = {
+  width: HEADER_ICON_SIZE,
```
cbt1, line 5: `HEADER_ICON_SIZE` is 12px but the icons are 13px. That is likely why the rendering tests are failing with "fuzzy" icons

a-m-dev: will update it soon

src/pivot-table/components/icons/styles.ts
```diff
@@ -0,0 +1,6 @@
+import { baseFlex } from "../shared-styles";
+
+export const headerPlusMinusIconWrapperStyle = {
+  width: 13,
```
cbt1, line 4: Maybe expose this pixel value as a constant and also consume it in the icon components?

cbt1 (approved):
Nice. Any idea as to why the icons scaled the way it did?

a-m-dev:
I think because `HEADER_ICON_SIZE` was 12 px and the width of icon we pass to it is 13px. 
so the svg icon will try to fit itself into the 12px wrapper div -> which will result in shrieked size

cbt1:
> I think because `HEADER_ICON_SIZE` was 12 px and the width of icon we pass to it is 13px. so the svg icon will try to fit itself into the 12px wrapper div -> which will result in shrieked size

I mean the "fix" part of this PR. Not the comment I had earlier.

cbt1 (approved):

src/pivot-table/components/shared-styles.ts
```diff
@@ -14,12 +14,16 @@ export enum Colors {
...
 export const CELL_PADDING = 4;
 
-export const topContainerCellStyle: React.CSSProperties = {
+export const baseFlex: React.CSSProperties = {
   display: "flex",
   justifyContent: "center",
   alignItems: "center",
 };
 
+export const topContainerCellStyle: React.CSSProperties = {
```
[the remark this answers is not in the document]
cbt1: I will remove that in my refactor dimension cells PR. As that styling is no longer needed at all

PR #430, opened 2023-10-23
feat: drag to resize columns in the header grid

haxxmaxx:
- Adding the column adjuster the to the header grid. Note that this will have to change a bit when the horizontal headers will be added
- creating a new type called `cellInfo` which is a subset of the `Cell` properties. We could probably align this a bit better with `HeaderCell`
- Adjusting the position and styling for the last column, both for the header grid and top grid


src/pivot-table/hooks/use-head-cell-dim.ts
```diff
@@ -8,7 +8,10 @@ interface UseHeadCellDim {
 export const useHeadCellDim = ({ interactions }: UseHeadCellDim) => {
   const [open, setOpen] = useState(false);
 
-  const handleOpenMenu = () => interactions.active && setOpen(true);
+  const handleOpenMenu = (evt: React.MouseEvent) =>
+    interactions.active &&
+    !(evt.target as HTMLElement)?.className.includes("sn-pivot-table-column-adjuster") &&
```
haxxmaxx, line 13: Potential refactor to have this as a util since we use it in multiple places now

src/pivot-table/data/extract-headers.ts
```diff
@@ -44,6 +44,7 @@ const extractHeaders = (
         qApprMaxGlyphCount: qDimensionInfo.qApprMaxGlyphCount,
         isDim: true,
         headTextAlign: "left",
+        isAncestorPseudoDimension: matrix[rowCount - 1].some((cell) => cell && !cell.isDim),
```
cbt1, line 47: This will be true no matter where in the "hierarchy" the pseudo dimension is position. So not really `isAncestorPseudoDimension` in that sense. But maybe that is the logic that is needed?

cbt1: Also consider declaring this so that it can be reused `const row = matrix[rowCount - 1];`

src/pivot-table/hooks/use-head-cell-dim.ts
cbt1: Should at least make a constant with the class name

src/pivot-table/data/extract-headers.ts
haxxmaxx: I don't think so. I loop over the entire row, since we first create the matrix containg null. But when the you create cell n, every cell > n is null. So if the pseudo dim hasn't been added yet, `isAncestorPseudoDimension` should be false

haxxmaxx: Or do you mean that it isn't really an "ancestor". Yeah that is true, since it is not a tree. I went with the same naming for consistency. I think it would make sense that `HeaderCell` and `Cell` have more common properties. using `x` and `y` instead of `colIdx` and `rowIdx` for example. But I know HeaderCell is aligned with the menu props. We would have to take a more thurough look at that if we want to rename things again

cbt1: Ah yes, nice 👍 

I do believe the time complexity is O(n^2). Where the worse case would be lots of dimensions with no pseudo dim. I doubt that is a problem but if so, you could probably store the index of the pseudo dim and compare that with the `colIdx` instead. That should then be O(n).

cbt1: Btw I think @a-m-dev could use `isAncestorPseudoDimension` in #437 to simplify the qDimensionInfo index calculation?

src/types/types.ts
```diff
@@ -98,6 +99,14 @@ export interface Cell {
   selectionCellType: NxSelectionCellType;
 }
 
+export interface CellInfo {
+  x?: number;
+  y?: number;
```
cbt1, line 104: I don't see that you assign `y` anywhere?

src/pivot-table/data/extract-headers.ts
a-m-dev: kind of, basically i just need an (actual) index of the dim which engine would interpret it correctly. 
So to be exact maybe not this variable of `isAncestorPseudoDimension` but the thing you mentioned that you would add to cell and I could then use to get the correct index without my calculations should be fine to make sort work correctly!

haxxmaxx:
>     1. (I'm not sure if we discussed this before or not, but) Total cells are not resizable:
Yeah I barely have tested with totals, still don't understand it fully either. I think that is something I will look in a different PR.
>     2. While resizing, the blue area can only move till available space to it's original cell size (i guess) but after releasing my mouse click, it resizes until the mouse position
So first off, I block the blue border when you can't make the cell smaller. I think it is 20px or something, arbitrary atm. So if you release when the pointer is to the left of that, the resulting widht is 20px

cbt1:
> > ```
> > 1. (I'm not sure if we discussed this before or not, but) Total cells are not resizable:
> > ```
> 
> Yeah I barely have tested with totals, still don't understand it fully either. I think that is something I will look in a different PR.

As of #432, totals should work. I will do some testing on it,

a-m-dev:
> > > ```
> > > 1. (I'm not sure if we discussed this before or not, but) Total cells are not resizable:
> > > ```
> > 
> > 
> > Yeah I barely have tested with totals, still don't understand it fully either. I think that is something I will look in a different PR.
> 
> As of #432, totals should work. I will do some testing on it.
> 
> EDIT: It works for me now. Could you retry @a-m-dev ?

Yes, Now it works and every measure cell reacts independently like below: 
Screenshot 2023-10-27 at 08 49 55

haxxmaxx:
> Yes, Now it works and every measure cell reacts independently like below: Screenshot 2023-10-27 at 08 49 55
> 
> which i think is the ideal behaviour here right? if so, LGTM 🚀

Yes, the width is always applied per dimension or measure. Also note that I haven't changed anything on the right side in this PR, but it is good that you are testing :)

src/pivot-table/hooks/use-head-cell-dim.ts
haxxmaxx: I'll leave this for the next PR. Will do some small refactoring

cbt1 (approved):

PR #433, opened 2023-10-24
fix: lock bg issue

a-m-dev:
the lock cell background is behaves differnetly in very big cell sizes (maybe over 30k px), this pr fixes it

in below examples i just stretch the cell width manually but i saw some nodes with very large children which i noticed the bug on it

Before: 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/3c57d127-5e00-4cb9-8362-176c59da651d

After: 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/340dece5-8457-4eb5-9a70-e64ce2859496

### DIFF
--- a/src/pivot-table/components/cells/utils/get-dimension-cell-style.ts
+++ b/src/pivot-table/components/cells/utils/get-dimension-cell-style.ts
@@ -63,6 +63,7 @@ export const getLockedStyleFromSelection = (originalBackgroundColor?: string): R
         ${originalBackgroundColor ?? Colors.Transparent} 0px 4px
       )`,
     color: "#bebebe",
+    backgroundClip: "padding-box",
   };
 };
 


cbt1 (approved):

cbt1:
Looks like this fix had some unexpected side effects on locked cells. Most noticable when a cell has some colorful background color. I will see if I can fix it in my #432 PR

Notice how the border colors changes to white when they are locked.

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/28748418-40df-4085-969f-0c7ca0fa6a11

cbt1:
The gradient issue only happens in Chrome. I do suspect that it's a browser issue. It "works" when you change `padding-box` to `border-box` but the gradient pattern will then render over the border. That artifact is barely visible since the border is only 1 px.

PR #436, opened 2023-10-25
fix: react border color warning

cbt1:
Fixes a react warning that could occur when switching themes. Reproduced by switching between "Default" and "Only transparent borders" theme in Nebula dev.

> Warning: Updating a style property during rerender (borderColor) when a conflicting property is set (borderBottomColor) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.

### DIFF
--- a/src/pivot-table/components/__tests__/shared-styles.test.ts
+++ b/src/pivot-table/components/__tests__/shared-styles.test.ts
@@ -17,13 +17,11 @@ describe("Shared styles", () => {
       borderRightWidth: 1,
       borderBottomWidth: 0,
       borderRightColor: borderColor,
-      borderBottomColor: undefined,
     };
     const borderBottom = {
       ...baseStyle,
       borderRightWidth: 0,
       borderBottomWidth: 1,
-      borderRightColor: undefined,
       borderBottomColor: borderColor,
     };
     const borderRightBottom = {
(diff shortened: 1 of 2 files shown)


src/pivot-table/components/__tests__/shared-styles.test.ts
```diff
@@ -4,11 +4,35 @@ import { borderStyle, cellStyle, getBorderStyle } from "../shared-styles";
...
     const noBorder = { ...baseStyle, borderRightWidth: 0, borderBottomWidth: 0 };
-    const borderRight = { ...baseStyle, borderRightWidth: 1, borderBottomWidth: 0 };
-    const borderBottom = { ...baseStyle, borderRightWidth: 0, borderBottomWidth: 1 };
-    const borderRightBottom = { ...baseStyle, borderRightWidth: 1, borderBottomWidth: 1 };
+    const borderRight = {
+      ...baseStyle,
+      borderRightWidth: 1,
+      borderBottomWidth: 0,
+      borderRightColor: borderColor,
+      borderBottomColor: undefined,
```
haxxmaxx, line 20: This is already undefined in baseStyle, so no need to override unless for readability

haxxmaxx (approved):

PR #437, opened 2023-10-25
fix: sort columns after pseudo dim in left grid

a-m-dev:
https://github.com/qlik-oss/sn-pivot-table/assets/29652890/4766f984-6739-491a-81ed-bb0de6da8b0c


cbt1 (approved):
Nice, I would think this will work out-of-the-box when we got headers for the top as well

PR #440, opened 2023-10-27
refactor: drag helpers

haxxmaxx:
- create `isEventFromColumnAdjuster` to block selections and opening header menu
- add a `canBeResized` prop on both `HeaderCell` and `Cell` that goes is the main parameter that determines if the adjuster renders or not
- move logic to not render `ColumnAdjuster` into itself. creates a cleaner look in the components it is used. But have to do the early return after I create a state and use some hooks. Not sure this is the best idea


src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -21,14 +22,18 @@ interface AdjusterProps {
...
   const tempWidth = useRef({ initWidth: 0, columnWidth: 0, initX: 0 });
   const positionAdjustment = isLastColumn ? CELL_PADDING : CELL_PADDING + GRID_BORDER;
+  const shouldRender = !isActive && cellInfo.canBeResized;
 
   useOnPropsChange(() => {
-    setInternalWidth(columnWidth);
+    if (shouldRender) setInternalWidth(columnWidth);
   }, [columnWidth]);
 
+  if (!shouldRender) return undefined;
```
cbt1, line 35: Return `null` when we don't want to render a componet. Not sure if React threats it differently by that is the value they use in the docs.

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -21,14 +22,18 @@ interface AdjusterProps {
  * While dragging this components follows the pointer, and on mouse up all column widths are updated.
  */
 const ColumnAdjuster = ({ cellInfo, columnWidth, dataModel, isLastColumn }: AdjusterProps) => {
+  const { isActive } = useSelectionsContext();
   const [internalWidth, setInternalWidth] = useState(columnWidth);
   const tempWidth = useRef({ initWidth: 0, columnWidth: 0, initX: 0 });
   const positionAdjustment = isLastColumn ? CELL_PADDING : CELL_PADDING + GRID_BORDER;
+  const shouldRender = !isActive && cellInfo.canBeResized;
```
cbt1, line 29: Should also include `interactions.active` in this check. So that you cannot re-size in edit mode in Sense.

haxxmaxx: yeah there is a fix PR comming up next, didn't want to mix it with this refactor

src/pivot-table/components/cells/ColumnAdjuster.tsx
haxxmaxx: yeah I was expecting a type error if it was incorrect, will change to null

src/pivot-table/components/cells/styles.tsx
```diff
@@ -13,6 +13,9 @@ interface StyledHeaderCellWrapperProps {
   isDimension: boolean;
 }
 
+export const COLUMN_ADJUSTER_CLASS = "sn-pivot-table-column-adjuster";
+const COLUMN_ADJUSTER_BORDER_CLASS = `& .${COLUMN_ADJUSTER_CLASS}-border`;
```
cbt1, line 17: export and consume in `ColumnAdjuster.tsx`?

cbt1: I guess the ampersand would have to be solved differently tho

cbt1 (approved):
Just minor comment. LGTM

src/pivot-table/components/cells/styles.tsx
haxxmaxx: yeah exactly, I can try though

PR #441, opened 2023-10-27
fix: unexpected scroll

a-m-dev:


src/pivot-table/components/cells/DimensionValue/index.tsx
```diff
@@ -53,6 +62,7 @@ const DimensionValue = ({ index, style, data }: DimensionValueProps): JSX.Elemen
       columnWidth={style.width as number}
       dataModel={dataModel}
       isLastColumn={isLastColumn}
+      scrollbarSharePerCol={parseFloat((verticalScrollbarWidth / layoutService.size.x).toFixed(12))}
```
haxxmaxx, line 65: I'm trying to understand why you need this. Is it needed because of your changes? Or was it a bug before? I think we should aim to make sure that the `ColumnAdjuster` receives a correct `columnWidth`, and not that it needs another argument to compensate for it.

a-m-dev: This is the current behaviour in main when the scrollbar is present. 
the columnWidth is up to date within Column adjuster component. the issue is that it does not consider the scrollbar width when it comes to calculate the position of the blue handle. 
https://github.com/qlik-oss/sn-pivot-table/assets/29652890/36a2a285-081b-4e70-997f-b414ef32127e

haxxmaxx: But it shouldn't need that. there is something else going wrong I think. The adjuster is positioned to using the width of the column - a some pixels to put it on the border. As long as the columnWidth passed to ColumnAdjuster is correct (which it should, it is the same value used to set the width of the column) and up to date, you shouldn't have to compensate for the scrollbar. That we already do when we calculate the column widths

haxxmaxx: I've tested a bit and it seems to be an intermittent issue. If the problem was not taking the scrollbar width into account, I'd expect the offset to be there all the time

a-m-dev: The unexpected scrollbar issue on main was because we were using `tableRect.width` and not `actualRightGridWidth` when there is **more space** than sum of col widths,

So after that change, I started to notice this issue and it seems like when you adjust a col width (when the scroll bar is **visible** and **takes space**), we are saving the actual column width and not the width added by the share of  the scrollbar width per column, so that is why the (blue/gray) handler is always a be a bit off the actual column width.

But that is my observation of the problem which i created this fix based on.

a-m-dev: this issue now scoped out to it's own verifail

src/pivot-table/components/PivotTable.tsx
```diff
@@ -137,8 +137,8 @@ export const StickyPivotTable = ({
           style={{
             display: "grid",
             gridTemplateColumns: leftGridWidth // If leftColumnsWidth is 0, this means no data exist for "headers" or "left"
-              ? `${leftGridWidth}px ${rightGridWidth}px`
-              : `${rightGridWidth}px`,
+              ? `${leftGridWidth}px ${rightGridWidth - verticalScrollbarWidth}px`
```
cbt1, line 140: Is this not something that should be substracted from in `get-scrollable-areas-dimensions`?

src/pivot-table/components/helpers/get-scrollable-areas-dimensions.ts
```diff
@@ -103,11 +106,11 @@ const getScrollableAreasDimensions = ({
           height: childWrappersScrollableContainerHeight,
         },
         fullSize: {
-          width: totalWidth - leftGridWidth - modifiedVerticalScrollbarWidth,
+          width: actualRightGridWidth,
           height: containerHeight,
         },
         sticky: {
-          width: tableRect.width - leftGridWidth - modifiedVerticalScrollbarWidth,
+          width: actualRightGridWidth < avaliableRightGridW ? actualRightGridWidth : avaliableRightGridW,
```
cbt1, line 113: Maybe a unit test that covers this scenario?

cbt1 (approved):
Just a minor comment. LGTM

PR #443, opened 2023-10-30
feat: update left side as you resize column

haxxmaxx:
- Stores `leftGridWidthInfo` as a state instead of useMemo, thus we have a setter that can update the widths on the fly
- export `overrideLeftGridWidth`, that is used in the `ColumnAdjuster` as you drag. Not super happy of how many layers we need to pass it through but don't think there is an alternative


src/pivot-table/hooks/use-column-width.ts
```diff
@@ -95,58 +98,79 @@ export default function useColumnWidth(
...
-      let width: number;
+      let sumOfWidths = 0;
+
+      const lastRow = headersData.data[headersData.size.y - 1] as HeaderCell[];
+      const columnWidths = lastRow.map((cell, index) => {
+        let width: number;
+
+        if (widthOverride && overrideIndex !== undefined && overrideIndex === index) {
+          width = widthOverride;
+        } else if (cell.id === PSEUDO_DIMENSION_KEY) {
```
haxxmaxx, line 123: This is the only change to the logic

src/pivot-table/hooks/__tests__/use-column-width.test.ts
```diff
@@ -84,17 +85,16 @@ describe("useColumnWidth", () => {
   const renderUseColumnWidth = () => {
     const {
       result: { current },
-    } = renderHook(() => {
-      headersData = createHeadersData(layoutService, visibleTopDimensionInfo, visibleLeftDimensionInfo);
```
haxxmaxx, line 88: this ended up creating a new rerender every time, creating an endless loop

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -96,57 +99,79 @@ export default function useColumnWidth(
...
+      qMeasureInfo,
+      qNoOfLeftDims,
+      rect.width,
+    ],
+  );
 
-      sumOfWidths += width;
-      return width;
-    });
+  const [leftGridWidthInfo, setLeftGridWidthInfo] = useState<LeftGridWidthInfo>(calculateLeftGridWidthInfo);
```
cbt1, line 171: This initial state is the function? Should this not be `useState<LeftGridWidthInfo>(calculateLeftGridWidthInfo());`?

haxxmaxx: no, that is not the syntax, quite confusing. If you pass a function to setState, it runs the function and sets the state to the return value. In order to set the state to function, you have to do `setState(() => nameOfFunction)`

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -96,57 +99,79 @@ export default function useColumnWidth(
...
-    const lastRow = headersData.data[headersData.size.y - 1] as HeaderCell[];
-    const columnWidths = lastRow.map((cell, index) => {
-      let width: number;
+      let sumOfWidths = 0;
+
+      const lastRow = headersData.data[headersData.size.y - 1] as HeaderCell[];
+      const columnWidths = lastRow.map((cell, index) => {
+        let width: number;
+
+        if (widthOverride && overrideIndex !== undefined && overrideIndex === index) {
```
a-m-dev, line 125:
```suggestion
        if (widthOverride && overrideIndex && overrideIndex === index) {
```

cbt1: Don't think that would work if `overrideIndex`is `0`

haxxmaxx: override index can be 0

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -156,8 +181,16 @@ export default function useColumnWidth(
...
     estimateWidthForDimensionValue,
+    calculateLeftGridWidthInfo,
   ]);
 
+  const overrideLeftGridWidth = useCallback(
+    (width: number, index: number) => {
+      setLeftGridWidthInfo(calculateLeftGridWidthInfo(width, index));
+    },
+    [calculateLeftGridWidthInfo],
+  );
```
a-m-dev, line 192: probably you dont need useCallback in this case?

src/pivot-table/hooks/use-column-width.ts
a-m-dev: maybe to prevent the confusion around this, it would be better to keep the () after the function name, if you look into this line and then look into line 174 in useEffect, definitely you will get confused

a-m-dev (approved):
LGTM + some minor comments

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: I'd say it is pretty basic hook knowledge ;) But I'll add it, no need for unnecessary confution

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: I didn't have the callback first, and there was something complaining or not working. I'll look at it again

haxxmaxx: that was for the `calculateLeftGridWidthInfo` function. So I think we could skip useCallback here. But I'd say it's better to take a complete look at this file. I think there is a overusage of useXxxx here, and we might want to break some things out as well. I think I prefer to keep this memoized, since we do it for everything else in this file. Especially the return values

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -97,59 +100,81 @@ export default function useColumnWidth(
...
 
-    return {
-      leftGridWidth: Math.min(rect.width * LEFT_GRID_MAX_WIDTH_RATIO, sumOfWidths),
-      leftGridColumnWidths: columnWidths,
-      leftGridFullWidth: sumOfWidths,
-    };
+  const [leftGridWidthInfo, setLeftGridWidthInfo] = useState<LeftGridWidthInfo>(calculateLeftGridWidthInfo());
+
+  useEffect(() => {
```
cbt1, line 176: Setting state in `useEffect`, use `useOnPropsChange` instead

cbt1:
As a side note for another PR. Probably want to have a different minimum size for columns on the left side as you can end with need seeing the text at all.

Skärmavbild 2023-11-03 kl  10 56 25

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -97,59 +100,81 @@ export default function useColumnWidth(
...
+      rect.width,
+    ],
+  );
 
-    return {
-      leftGridWidth: Math.min(rect.width * LEFT_GRID_MAX_WIDTH_RATIO, sumOfWidths),
-      leftGridColumnWidths: columnWidths,
-      leftGridFullWidth: sumOfWidths,
-    };
+  const [leftGridWidthInfo, setLeftGridWidthInfo] = useState<LeftGridWidthInfo>(calculateLeftGridWidthInfo());
```
cbt1, line 174: Did you end up calling the function anyway?

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: I was wondering about this, thanks

haxxmaxx:
> As a side note for another PR. Probably want to have a different minimum size for columns on the left side as you can end with not seeing the text at all.

Yeah the min width is completely arbitrary at this point, and static. We probably want to set it to something more intentional and make it depend on font size

src/pivot-table/hooks/use-column-width.ts
cbt1: It's actually 2 different uses. By doing `useState<LeftGridWidthInfo>(calculateLeftGridWidthInfo);` the function is only called the first the time the hook is "rendered". With `useState<LeftGridWidthInfo>(calculateLeftGridWidthInfo());` it's called on every render, which is probably not what we want, nor is it going to be better from a performance spective.

IMO change it back and add a comment instead.

I will try and do some quick testing on the impact of it.

cbt1: Some quick testing: `calculateLeftGridWidthInfo` is called 4 times as much now compared to before when re-sizing a column

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: yes, to avoid the confusion

cbt1: See my comment about the performance impact of this

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: fixed

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: oh, I didn't see this discussion. I'll change it back :)

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: fixed

cbt1 (approved):
🚢

PR #444, opened 2023-10-30
feat: hover on expand icon

cbt1:
EXPERIMENTAL

Trying to make it easier to hit the expand/collapse icon. Currently the icon is highlighted by changing background on hover. But that does not really make all that much easier to hit.


cbt1:
N/A

PR #445, opened 2023-10-31
fix: text line clamp for expanded cell

cbt1:
Allows cell that are expanded to have text on multiple lines.

marginTop is need on left side to put the Icon aligned at the center of the first row. Previously this was handled by `alignItems: center` in `StickCellContainer.tsx`. Now it have to be manually positioned.

Before:
Skärmavbild 2023-10-31 kl  11 52 17

After:
Skärmavbild 2023-10-31 kl  11 52 01

### DIFF
--- a/src/pivot-table/components/cells/DimensionValue/ExpandOrCollapseIcon.tsx
+++ b/src/pivot-table/components/cells/DimensionValue/ExpandOrCollapseIcon.tsx
@@ -24,6 +24,8 @@ interface OnClickHandlerProps {
 export const testIdExpandIcon = "expand-icon";
 export const testIdCollapseIcon = "collapse-icon";
 
+const halfIconHeight = PLUS_MINUS_ICON_SIZE / 2;
+
 const createOnClickHandler =
   ({ expandOrCollapse, cell }: OnClickHandlerProps) =>
   (e: React.SyntheticEvent) => {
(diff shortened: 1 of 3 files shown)


src/pivot-table/components/cells/DimensionValue/Text.tsx
```diff
@@ -16,7 +16,11 @@ const Text = ({ children, cell, styleService, isCellSelected, isLeftColumn }: Pr
   <span
     style={{
       ...textStyle,
-      ...getLineClampStyle(isLeftColumn ? styleService.grid.lineClamp : DEFAULT_LINE_CLAMP),
+      ...getLineClampStyle(
+        isLeftColumn
+          ? Math.max(styleService.grid.lineClamp, cell.leafCount * styleService.grid.lineClamp)
```
cbt1, line 21: When leafCount is 0 fallback to the styleservice value

cbt1: Note: This logic is very much simplified. To ensure that the whole available cell height is used, the below logic would need to be used. Plus handle cell padding.

```js
const getLineClamp = ({ isLeftColumn, cell, styleService }: Pick<Props, "cell" | "isLeftColumn" | "styleService">) => {
  if (isLeftColumn) {
    if (cell.ref.qCanCollapse) {
      const textHeight = parseInt(styleService.dimensionValues.fontSize, 10) * LINE_HEIGHT_COEFFICIENT;
      const cellHeight = Math.max(styleService.contentCellHeight, cell.leafCount * styleService.contentCellHeight);

      return Math.floor((cellHeight / textHeight) * styleService.grid.lineClamp);
    }

    return styleService.grid.lineClamp;
  }

  return DEFAULT_LINE_CLAMP;
};
```

src/pivot-table/components/cells/DimensionValue/ExpandOrCollapseIcon.tsx
```diff
@@ -23,6 +24,8 @@ interface OnClickHandlerProps {
 export const testIdExpandIcon = "expand-icon";
 export const testIdCollapseIcon = "collapse-icon";
 
+const halfIconHeight = PLUS_MINUS_ICON_SIZE / 2;
```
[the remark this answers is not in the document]
cbt1: Will change in next PR

PR #447, opened 2023-10-31
fix: column adjuster handle 

a-m-dev:
here is the sum of what we discussed this morning about col adjuster, thanks @cbt1 for this


src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -14,47 +13,41 @@ interface AdjusterProps {
...
-
 /**
  * Component that is placed on top of column border, to resize the columns.
  * When you start dragging, mouse move and mouse up listeners are added.
  * While dragging this components follows the pointer, and on mouse up all column widths are updated.
  */
 const ColumnAdjuster = ({ cellInfo, columnWidth, dataModel, isLastColumn }: AdjusterProps) => {
-  const [internalWidth, setInternalWidth] = useState(columnWidth);
-  const tempWidth = useRef({ initWidth: 0, columnWidth: 0, initX: 0 });
+  const [, setInternalWidth] = useState(columnWidth);
```
a-m-dev, line 22: the only use case for this is just triggering update re-render then?

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -14,47 +13,41 @@ interface AdjusterProps {
...
   const positionAdjustment = isLastColumn ? CELL_PADDING : CELL_PADDING + GRID_BORDER;
 
-  useOnPropsChange(() => {
-    setInternalWidth(columnWidth);
-  }, [columnWidth]);
```
a-m-dev, line 30: where this could be useful? any reason for not removing it?

src/pivot-table/components/cells/ColumnAdjuster.tsx
haxxmaxx: you mean the the internalWidth state?

a-m-dev: Yes

cbt1, line 22: Maybe rename `setInternalWidth` to `forceRerender` or something like that and just set the value to a new object when we need it to rerender? As I think it can be confussing to keep the column width value here but not actually using it.

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -14,47 +13,41 @@ interface AdjusterProps {
...
   const mouseDownHandler = (evt: React.MouseEvent) => {
     evt.stopPropagation();
 
-    tempWidth.current = {
-      initX: evt.clientX,
-      initWidth: columnWidth,
-      columnWidth,
-    };
+    tempWidth.initX = evt.clientX;
+    tempWidth.initWidth = columnWidth;
```
cbt1, line 49: I don't think you need to assing `initWidth` and `columnWidth` if you set the values via useMemo instead:
`const tempWidth = useMemo(() => ({ initWidth: columnWidth, columnWidth, initX: 0 }), [columnWidth]);`

a-m-dev: we cant create  hook within the nested function so we cant do it

src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -14,47 +13,41 @@ interface AdjusterProps {
...
 const ColumnAdjuster = ({ cellInfo, columnWidth, dataModel, isLastColumn }: AdjusterProps) => {
-  const [internalWidth, setInternalWidth] = useState(columnWidth);
-  const tempWidth = useRef({ initWidth: 0, columnWidth: 0, initX: 0 });
+  const [, setInternalWidth] = useState(columnWidth);
   const positionAdjustment = isLastColumn ? CELL_PADDING : CELL_PADDING + GRID_BORDER;
 
-  useOnPropsChange(() => {
-    setInternalWidth(columnWidth);
-  }, [columnWidth]);
+  const tempWidth = useMemo(() => ({ initWidth: 0, columnWidth, initX: 0 }), [columnWidth]);
```
cbt1, line 25: Change this to `const tempWidth = useMemo(() => ({ initWidth: columnWidth, columnWidth, initX: 0 }), [columnWidth]);` and you dont need to assign values in the `mouseDownHandler`.
```js
// These can be removed
tempWidth.initWidth = columnWidth;
 tempWidth.columnWidth = columnWidth;
```

src/pivot-table/components/cells/ColumnAdjuster.tsx
cbt1: See my other comment about this

cbt1 (approved):
LGTM

PR #448, opened 2023-11-01
fix: text margin top

cbt1:
Fixes an issue where the margin top applied on the Text was not correct when chart was configured with a `lineClamp` greater then 1.

Also fixes an issue where the height was not correctly calculated for a single row when measure values had a custum font size configured that was larger then the font size for dimension values.

Before:
Skärmavbild 2023-11-01 kl  11 55 08

After:
Skärmavbild 2023-11-01 kl  11 54 54

### DIFF
--- a/src/services/style-service.ts
+++ b/src/services/style-service.ts
@@ -238,6 +238,10 @@ const createStyleService = (theme: ExtendedTheme, layoutService: LayoutService):
     DEFAULT_CELL_HEIGHT,
   );
 
+  styleService.contentRowHeight = styleService.contentCellHeight / lineClamp;
+
+  styleService.contentTextHeight = styleService.contentRowHeight - CELL_PADDING_HEIGHT;
+
   return styleService;
 };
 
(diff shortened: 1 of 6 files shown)


src/services/style-service.ts
```diff
@@ -238,6 +238,10 @@ const createStyleService = (theme: ExtendedTheme, layoutService: LayoutService):
     DEFAULT_CELL_HEIGHT,
   );
 
+  styleService.contentRowHeight = styleService.contentCellHeight / lineClamp;
```
a-m-dev, line 241:
```suggestion
  styleService.contentRowHeight = (styleService.contentCellHeight - CELL_PADDING_HEIGHT) / lineClamp;
```
shouldn't this give you the correct row height?
and then the `styleService.contentTextHeight` would be the same as `styleService.contentRowHeight` and wont be needed any more?

cbt1: Both are needed in `src/pivot-table/components/cells/DimensionValue/Text.tsx:16`

a-m-dev: the thing that got me thinking was that do we add padding to **each row** or for **each cell**? 
if it is cell -> then i think we should subtract it like the suggestion, right?

cbt1: Padding is added once to `contentCellHeight`. Even if `lineClamp` is `10`. `contentRowHeight` in a cell is the rows in which text can be draw at. So the calculations should be correct.

Note that a cell that is expanded can have a rendered "cell height" which is `contentCellHeight * <number-of-leaf-nodes>`. But for the context of these values, that does not matter.

a-m-dev (approved):

PR #450, opened 2023-11-02
fix: handle no theme border value

cbt1:
Fixes an issue where the resolved value from the `theme.getStyle()` would be an object if the active theme did not have any values set for `object.pivotTableV2.grid.border`. This would result in borders getting the wrong colors.

Before
image


a-m-dev (approved):

cbt1:
N/A anymore

PR #452, opened 2023-11-02
fix: dont render columnadjuster when interactions.active is false

haxxmaxx:


src/pivot-table/components/cells/ColumnAdjuster.tsx
```diff
@@ -20,10 +21,11 @@ interface AdjusterProps {
  * While dragging this components follows the pointer, and on mouse up all column widths are updated.
  */
 const ColumnAdjuster = ({ cellInfo, columnWidth, dataModel, isLastColumn }: AdjusterProps) => {
+  const { interactions } = useBaseContext();
   const { isActive } = useSelectionsContext();
   const [, forceRerender] = useState({});
   const positionAdjustment = isLastColumn ? CELL_PADDING : CELL_PADDING + GRID_BORDER;
-  const shouldRender = !isActive && cellInfo.canBeResized;
+  const shouldRender = cellInfo.canBeResized && !!interactions.active && !isActive;
```
haxxmaxx, line 28: I just moved these around a bit as well. Checking the most important bool first, so we don't do more than one check for the cells that will never have a column adjuster

cbt1 (approved):
🚢

PR #458, opened 2023-11-13
fix: crash when no data exists

cbt1:
Fixes an issue where the chart would crash if no data existed. This could for example occur after a selection in another object or by using some combination of properties that resulted in no data (easily trigged by using `qSuppressZero: true` and measure with only 0 values OR using `qSuppressMissing: true` and measure with only null values).

Now renders a disclaimer text in the middle of the chart. This is consistent with how other charts in Sense handles this.
Skärmavbild 2023-11-13 kl  14 20 59


src/pivot-table/hooks/use-disclaimer.ts
```diff
@@ -0,0 +1,39 @@
+import type { stardust } from "@nebula.js/stardust";
+import type { LayoutService } from "../../types/types";
+import { DEFAULT_FONT_FAMILY, DISCLAIMER_HEIGHT } from "../constants";
+import { useBaseContext } from "../contexts/BaseProvider";
+
+export interface DisclaimerProps {
+  translator: stardust.Translator;
+  layoutService: LayoutService;
+}
```
a-m-dev, line 9: I think this belongs to component not hook, because it originates from component and the hook is the consumer of that prop types

a-m-dev (approved):

src/pivot-table/hooks/use-disclaimer.ts
cbt1: Agree

PR #459, opened 2023-11-14
test: fix theme styling in rendering tests

cbt1:
Fixes an issue in the rendering tests where the "correct" styling was not used because when resolving the theme styling, it would fallback to the base values for some attributes (ex: background and color). This would render null values cells with transparent background.

- Added the default theme as part of the nebula serve config
- Added a nebula configure file for rendering tests
- Added script to debug rendering tests locally

Expected changes in rendering test baseline images
- Font size
- Font family


package.json
```diff
@@ -80,6 +80,9 @@
     },
     "root": true
   },
+  "resolutions": {
+    "rollup": "npm:rollup@3.29.4"
```
cbt1, line 84: Temporary solution that fixes two issues.
1. Rollup v4 installs a binary based on `os`and `cpu`. That binary is used as part of nebulas build/serve command. But it might not match the `os/cpu` when running `nebula serve` command inside the docker container through `yarn test:local:rendering` or `yarn test:local:update:screenshots`.
2. The "watch" functionally of nebula is broken since nebula updated to rollup v4. So it does not pick up code changes.

enell (approved):

PR #460, opened 2023-11-14
fix: prevent click event when dragging

haxxmaxx:
In order to block click events on the dimension value cells and header cells, we need to know when the column adjuster is being dragged or not. Therefor we add a `isAdjustingWidth` state, the setter to the `ColumnAdjuster` and set it to true when you start dragging. The state is reset if there is a layout change, so you can select/open the header menu again. Note that there is still a small bug you can get. If you have the min width aldready, then start to drag over the cell and release, you will trigger the cell click handler. This is an extreme corner-case, that will be fixed if we implement drag to select. See both the fix and the remaining bug in the gif
Kapture 2023-11-14 at 15 53 29

[This](https://github.com/qlik-oss/nebula-table-utils/pull/76) change in utils repo needs to be released before merging this PR

### DIFF
--- a/src/pivot-table/hooks/__tests__/use-head-cell-dim.test.ts
+++ b/src/pivot-table/hooks/__tests__/use-head-cell-dim.test.ts
@@ -2,18 +2,21 @@ import type { stardust } from "@nebula.js/stardust";
 import { COLUMN_ADJUSTER_CLASS } from "@qlik/nebula-table-utils/lib/constants";
 import { renderHook } from "@testing-library/react";
 import { act } from "react-dom/test-utils";
+import type { DataModel } from "../../../types/types";
 import { useHeadCellDim } from "../use-head-cell-dim";
 
 describe("UseHeadCellDim", () => {
   let interactions: stardust.Interactions;
   let evt: React.MouseEvent;
+  let dataModel: DataModel;
 
   beforeEach(() => {
     interactions = { active: true };
     evt = {} as React.MouseEvent;
+    dataModel = {} as DataModel;
   });
 
-  const renderer = () => renderHook(() => useHeadCellDim({ interactions })).result;
+  const renderer = () => renderHook(() => useHeadCellDim({ interactions, dataModel })).result;
 
   test("should return initial open state", () => {
     const result = renderer();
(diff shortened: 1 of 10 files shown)


src/pivot-table/hooks/__tests__/use-head-cell-dim.test.ts
```diff
@@ -52,6 +55,20 @@ describe("UseHeadCellDim", () => {
     expect(result.current.open).toBe(false);
   });
 
+  test("handleOpenMenu should not set open state to true when isAdjustingWidth is true", async () => {
+    const result = renderer();
+
+    await act(async () => {
```
haxxmaxx, line 61: I tried without the async here but it did not work when you have two act in a row

src/pivot-table/hooks/use-head-cell-dim.ts
```diff
@@ -1,16 +1,26 @@
...
 
 interface UseHeadCellDim {
   interactions: stardust.Interactions;
+  dataModel: DataModel;
 }
 
-export const useHeadCellDim = ({ interactions }: UseHeadCellDim) => {
+export const useHeadCellDim = ({ interactions, dataModel }: UseHeadCellDim) => {
   const [open, setOpen] = useState(false);
+  const [isAdjustingWidth, setIsAdjustingWidth] = useState(false);
```
cbt1, line 14: Does it have to be a state? Can it not be a reference?

haxxmaxx: yeah I tried with a ref and it doesn't work for selections, but it works for opening the menu... Not sure why and optimally we should probably know why. the state works at least and I can't see how this would be a big performance impact

src/pivot-table/hooks/use-head-cell-dim.ts
```diff
@@ -1,16 +1,26 @@
...
+
+  // dataModel is only needed here since we want to reset isAdjustingWidth whenever there is a layout change
```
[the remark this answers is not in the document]
haxxmaxx: this is still relevant, just moved the hook

src/pivot-table/hooks/__tests__/use-head-cell-dim.test.ts
cbt1: Not that I think it will matter but `act` should be imported from `import { act } from "@testing-library/react";`

cbt1 (approved):
LGTM

PR #462, opened 2023-11-16
fix: only render columnadjuster for headers that are left dimensions + correctly override widths on the left side

haxxmaxx:
- don't render the column adjuster for row headers
- set `colIdx` and use it for `overrideLeftGridWidth`. Now `colIdx` is used to set `activelySortedColumn`, which in turn determines some states in the header menu. That is still not correct and needs to be fixed in a different PR. This is still an improvement of what it was before.

I tried to test this, but to get a correct `isLeftDimension`, we need to mock more or less all of `layoutService`. We might want to do that in a smart way, but that is a different PR

### DIFF
--- a/src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
+++ b/src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
@@ -35,7 +35,7 @@ const ColumnAdjusterWrapper = ({
 
   const updateWidth = (adjustedWidth: number) => {
     forceRerender({});
-    if (overrideLeftGridWidth && cellInfo.colIdx) {
+    if (overrideLeftGridWidth && cellInfo.colIdx !== undefined) {
       overrideLeftGridWidth(adjustedWidth, cellInfo.colIdx);
     }
   };
--- a/src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
+++ b/src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
@@ -35,7 +35,9 @@ const ColumnAdjusterWrapper = ({
 
   const updateWidth = (adjustedWidth: number) => {
     forceRerender({});
-    overrideLeftGridWidth?.(adjustedWidth, cellInfo.dimensionInfoIndex);
+    if (overrideLeftGridWidth && cellInfo.colIdx) {
+      overrideLeftGridWidth(adjustedWidth, cellInfo.colIdx);
+    }
   };
 
   const confirmWidth = (newWidthData: ColumnWidth) => dataModel?.applyColumnWidth(newWidthData, cellInfo);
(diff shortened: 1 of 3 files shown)


src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
```diff
@@ -35,7 +35,9 @@ const ColumnAdjusterWrapper = ({
 
   const updateWidth = (adjustedWidth: number) => {
     forceRerender({});
-    overrideLeftGridWidth?.(adjustedWidth, cellInfo.dimensionInfoIndex);
+    if (overrideLeftGridWidth && cellInfo.colIdx) {
```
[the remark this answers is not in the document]
haxxmaxx: good catch. it is optional so need to sanity check it. I could cast it since if overrideLeftGridWidth exists, so does colIdx, but I think it is cleaner with an if statement

src/pivot-table/data/extract-headers.ts
```diff
@@ -113,11 +104,13 @@ const extractHeaders = (
...
-  if (matrix.length > 0) {
-    for (const cell of matrix[matrix.length - 1]) {
+  // Update canBeResized, isActivelySorted and colIdx on bottom row
+  if (matrix.length) {
+    for (const [colIdx, cell] of matrix[matrix.length - 1].entries()) {
       if (cell) {
-        cell.canBeResized = true;
+        cell.canBeResized = cell.isLeftDimension;
+        cell.isActivelySorted = colIdx === (layoutService.layout.qHyperCube.activelySortedColumn?.colIdx ?? 0);
+        cell.colIdx = cell.colIdx || colIdx;
```
cbt1, line 113: Will be reassigned if `cell.colIdx === 0`

```suggestion
        cell.colIdx = cell.colIdx ?? colIdx;
```

src/pivot-table/data/extract-headers.ts
```diff
@@ -53,21 +44,21 @@ const createHeaderCell = (
   }
   return {
     id,
-    colIdx,
+    colIdx: 0, // has to be properly set later, would be incorrect after transposing otherwise
```
cbt1, line 47: If this is set to `0` here, does that mean that cells at last column but not at the bottom row are all getting `colIdx === 0`? Since only the last row is reassigned later?

src/pivot-table/data/extract-headers.ts
haxxmaxx: not really actually. Since this is set by default to 0, it should actually change when it is 0. It is only fore pseudo dim it should not be updated

src/pivot-table/data/extract-headers.ts
haxxmaxx: yeah that is what I try to explain in the PR description. This value was incorrect before, both for the header menu and the column adjuster. Now it is correct for the column adjuster at least. This needs another look

cbt1 (approved):

PR #434, opened 2023-10-24
feat: add horizontal headers

enell:
Refs: VIZ-2530

PR #463, opened 2023-11-17
fix: last cell width overrides when specified

haxxmaxx:
The default behavior when you have horizontal headers is that you take the max width of the stacked headers. This change makes sure that if you have **specified a non-auto width for the last bottom cell, that will override the other horizontal headers**.

Kapture 2023-11-21 at 10 00 14


src/pivot-table/hooks/__tests__/use-column-width.test.ts
```diff
@@ -352,7 +352,7 @@ describe("useColumnWidth", () => {
       layoutService.layout.qHyperCube.qMeasureInfo = [meaInfo, meaInfoPixels, meaInfoFitToContent];
 
       const { getRightGridColumnWidth } = renderUseColumnWidth();
-      expect(getRightGridColumnWidth(0)).toBe(98);
+      expect(getRightGridColumnWidth(0)).toBe(132);
```
haxxmaxx, line 355: Now that the left column width is determined by the bottom row again, resetting this value to what it was before this https://github.com/qlik-oss/sn-pivot-table/pull/434/files#diff-f85cc6b2e82b4ca1f93f5d8293113f571eb9b79d2540e85ce634d378e7ba28e3R355

src/pivot-table/hooks/use-column-width.ts
```diff
@@ -158,6 +155,12 @@ export default function useColumnWidth(
               cellWidth = getColumnWidth(columnWidth, fitToContentWidth);
             }
 
+            // The last cell setting should override the other cells in that column, so we don't pick the max
+            const isTypeAuto = !header.columnWidth || header.columnWidth.type !== ColumnWidthType.Auto;
```
cbt1, line 159: Seems like this variable should be called `isNotAuto`?

cbt1:
> The default behavior when you have horizontal headers is that you take the max width of the stacked headers. This change makes sure that if you have specified a non-auto width for the last bottom cell, that will override the other horizontal headers.

This does not seem to be the case when I looked at the failed rendering test? As the icon for the "top" stacked header have dissappared.

src/pivot-table/hooks/use-column-width.ts
haxxmaxx: this was incorrect, which also caused the rendering test failure

cbt1 (approved):

PR #468, opened 2023-11-21
fix: use colIdx on header to override pseudo dim width

haxxmaxx:
Since the `overrideIndex` for pseudo dim is -1, we need to look at the `colIdx` on the header, not the index for the array


cbt1 (approved):
LGTM not sure I fully understood the issue tho

haxxmaxx:
> LGTM not sure I fully understood the issue tho

`overrideIndex === collIdx` was never true for pseudo dim, since it colIdx is the itterator of the array, so it is always > -1

PR #472, opened 2023-11-22
fix: resize flickering

cbt1:
Fixes an issue where the chart would be flickering when it was being resized.

Before:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/b5f88f4f-98ff-4242-98c3-807f9c953323

After:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/8e1720b0-d0a5-4b96-ad3f-50dfe201237f


haxxmaxx (approved):
clever 👍

PR #475, opened 2023-11-22
refactor: split up use-column-width

haxxmaxx:
Creating a new folder with all the use-column-width code. Breaking out two new hooks. One for the left side stuff and one for the right side. Naming can be discussed. The other stuff remains in the common index file. I've made some slight changes to some utils.
- `getPercentageValue` takes another argument to actually return the value that we will apply
- `getMeasureTextArgs` added to reduce some code when using measure text hook
- extracting `getColumnWidthValue` since it is pure with one argument added.


src/pivot-table/hooks/use-column-width/use-column-width-left.ts
```diff
@@ -0,0 +1,141 @@
...
+    isFullyExpanded,
+  } = layoutService;
+  const styleService = useStyleContext();
+  const { measureText: measureTextForHeader } = useMeasureText(getMeasureTextArgs(styleService.header));
+  const { measureText: measureTextForMeasureValue } = useMeasureText(getMeasureTextArgs(styleService.measureValues));
+  const { estimateWidth: estimateWidthForDimensionValue, measureText: measureTextForDimensionValue } = useMeasureText(
+    getMeasureTextArgs(styleService.dimensionValues),
+  );
+
+  const calculateLeftGridWidthInfo = useCallback(
```
haxxmaxx, line 31: I tried breaking some logic out of this one, but it is all very connected, and you would need to pass a lot of args

src/pivot-table/hooks/use-column-width/utils.ts
```diff
@@ -0,0 +1,30 @@
+import type { ColumnWidth } from "@qlik/nebula-table-utils/lib/components/ColumnAdjuster";
```
haxxmaxx, line 1: I learned recently that `utils` is usually used as a name for code that is more of a tool, like a debugging tool or something, not just extracted logic. Maybe this should be called something different? helpers?

cbt1: Never heard that before. But since we use `utils` I think we should continue with that for now. If we like to it would probably be better to change "all" at once to something like "helper".

cbt1 (approved):
LGTM. I have not looked in detail at the code since it's mostly just moving code.

My only comment is that there is likely an oppertunity to break out even more code (later) into smaller hooks to make it less complex to test. Also everything is still tested in a single file, I would think breaking that into smaller test files also might help handling all dependencies and get a better overview of all scenarios.

But never the less a good change 👍

haxxmaxx:
> LGTM. I have not looked in detail at the code since it's mostly just moving code.
> 
> My only comment is that there is likely an oppertunity to break out even more code (later) into smaller hooks to make it less complex to test. Also everything is still tested in a single file, I would think breaking that into smaller test files also might help handling all dependencies and get a better overview of all scenarios.
> 
> But never the less a good change 👍

Well, I kind of took the approach that you test the "end-point" of the hook, so it is independent to refactors like this. But maybe it is a bit too wide. I don't think you need to test the utils file seperately, but rather testing the resulting returns from using the hook in different scenarios. But yeah maybe we should split the tests up into smaller files, and then keep making smaller hooks, if possible

src/pivot-table/hooks/use-column-width/utils.ts
haxxmaxx: I took a couse on codacademy, and they explained it in that way. Googling it again it is kind of vague, so I agree lets not change, especially for consistency

cbt1:
> > LGTM. I have not looked in detail at the code since it's mostly just moving code.
> > My only comment is that there is likely an oppertunity to break out even more code (later) into smaller hooks to make it less complex to test. Also everything is still tested in a single file, I would think breaking that into smaller test files also might help handling all dependencies and get a better overview of all scenarios.
> > But never the less a good change 👍
> 
> Well, I kind of took the approach that you test the "end-point" of the hook, so it is independent to refactors like this. But maybe it is a bit too wide. I don't think you need to test the utils file seperately, but rather testing the resulting returns from using the hook in different scenarios. But yeah maybe we should split the tests up into smaller files, and then keep making smaller hooks, if possible

Yeah I yet to figure out when to make that kind of seperation. On one hand I think having it more at the "end-point" as you said, does for example allow for refactoring more easily without having to worry about having to re-write the tests or have some become redundent. On the other hand, having large complex test suit will increase the likelyhood of bugs in the tests and can make the connection between input and output harder to understand.

Maybe a healthy mix between the two approaches can yield the best result?

haxxmaxx:
> Yeah I yet to figure out when to make that kind of seperation. On one hand I think having it more at the "end-point" as you said, does for example allow for refactoring more easily without having to worry about having to re-write the tests or have some become redundent. On the other hand, having large complex test suit will increase the likelyhood of bugs in the tests and can make the connection between input and output harder to understand.
> 
> Maybe a healthy mix between the two approaches can yield the best result?

I had a quick look at the test file. and the problem I would say is that some of the props we return depend on both the left and right side. So if you split thing up in different files, you would have to mock a lot of things multiple times. There is surely a better way to structure this, but would require some thought. Might get into that if I have some time over. I will merge this for now

PR #479, opened 2023-11-25
feat: adjust last header column when values are in rows

enell:
Before:

https://github.com/qlik-oss/sn-pivot-table/assets/968450/9b750407-c094-41b1-a417-403f7bfb58e6

After:

https://github.com/qlik-oss/sn-pivot-table/assets/968450/93881b6c-6ed5-49df-b81d-a1e31652e27a


haxxmaxx:
So tested a bit. The main problem with this is that there is no way of setting the width of the column now, that is smaller than the fitToContent value for `year`. I'm pretty sure some user wants to make the chart as compact as possible. It becomes even more appreant in this case
Screenshot 2023-11-27 at 10 52 50
You get a lot of white space next to the cylinder values, that you can't reduce (or only from the cylinders header)

Not sure what we want to do here, maybe we need to run it by UX?

src/pivot-table/data/extract-headers.ts
```diff
@@ -112,13 +118,24 @@ const extractHeaders = (
...
-        cell.canBeResized = cell.isLeftDimension;
         cell.isActivelySorted = colIdx === (layoutService.layout.qHyperCube.activelySortedColumn?.colIdx ?? 0);
         cell.colIdx = cell.colIdx || colIdx;
+
+        if (cell.dimensionInfoIndex === PSEUDO_DIMENSION_INDEX) {
+          cell.columnWidthLocation = ColumnWidthLocation.Measures;
+        } else if (!cell.isLeftDimension) {
+          if (visibleLeftDimensionInfos.at(-1) === PSEUDO_DIMENSION_INDEX) {
+            cell.columnWidthLocation = ColumnWidthLocation.Measures;
+          } else if (visibleLeftDimensionInfos.indexOf(PSEUDO_DIMENSION_INDEX) !== -1) {
```
cbt1, line 133: Looks like this is what you are trying to evaluate?

```suggestion
          } else if (layoutService.hasPseudoDimOnLeft) {
```

src/types/types.ts
```diff
@@ -16,13 +16,19 @@ export type List = Record<number, Cell>;
 
 export type Grid = List[];
 
+export enum ColumnWidthLocation {
+  Dimension = "dimensions",
+  Measures = "measures",
```
cbt1, line 21: Why not name it `PseudoDimensions`?

src/pivot-table/hooks/use-data-model.ts
```diff
@@ -82,43 +84,45 @@ export default function useDataModel({
...
+          paths = [
+            {
+              qPath: `${Q_PATH}/qDimensions/${dimensionInfoIndex}/qDef/columnWidth`,
+              oldValue: qDimensionInfo[dimensionInfoIndex].columnWidth,
+            },
+          ];
+          break;
+        case ColumnWidthLocation.Measures:
+          if (isLeftColumn) {
+            paths = [...Array(qMeasureInfo.length).keys()].map((idx) => ({
```
cbt1, line 102: I would think this will give the same output

```suggestion
            paths = qMeasureInfo.map((info, idx) => ({
```

src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
```diff
@@ -32,7 +32,7 @@ const ColumnAdjusterWrapper = ({
   const { isActive } = useSelectionsContext();
   const styleService = useStyleContext();
   const [, forceRerender] = useState({});
-  const shouldRender = cellInfo.canBeResized && !!interactions.active && !isActive;
```
haxxmaxx, line 35: We still need this. This component handles all the logic for whether it should be rendered or not. After this change, all dimensionValue and header cells have a column adjuster

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
```diff
@@ -45,47 +73,34 @@ export default function useColumnWidthLeft({ layoutService, tableRect, headersDa
...
             }
-
-            // The last cell setting should override the other cells in that column, so we don't pick the max
-            const isTypeAuto = !header.columnWidth || header.columnWidth.type === ColumnWidthType.Auto;
-            if (lastRowLastColumn && !isTypeAuto) {
-              return cellWidth;
+          } else if (!lastRowHeader.isLeftDimension) {
+            fitToContentWidth = dimensionsFitToContentWidth(collIdx);
+            if (lastRowHeader.columnWidthLocation === ColumnWidthLocation.Measures) {
+              fitToContentWidth = measureFitToContentWidth(fitToContentWidth);
```
haxxmaxx, line 96: any reason doing not doing this as a if ... else?

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
```diff
@@ -31,10 +31,38 @@ export default function useColumnWidthLeft({ layoutService, tableRect, headersDa
...
-        const fitToContentWidth = measureTextForMeasureValue(qFallbackTitle) + TOTAL_CELL_PADDING;
-        return Math.max(maxWidth, getColumnWidthValue(tableRect.width, columnWidth, fitToContentWidth));
-      }, 0);
+      const dimensionHeaderCellWidth = (lastRowHeader: HeaderCell) => {
+        const { label, isLocked } = lastRowHeader;
+        const lockedIconSize = isLocked ? LOCK_ICON_SIZE : 0;
+        return TOTAL_CELL_PADDING + measureTextForHeader(label) + MENU_ICON_SIZE + lockedIconSize;
+      };
+
+      const measureFitToContentWidth = (fitToContentWidth = 0) =>
```
haxxmaxx, line 40: is this really the fitToContent width? you apply a value as argument, then call getColumnWidthValue. I see why you want to get it as that, but this would be the final width for a column, when you take all measures into account

src/types/types.ts
enell: I wanted this to reflect where in hyper cube the column width were stored. So measures means, `/qHyperCubeDef/qMeasures/${idx}/qDef/columnWidth`. `PseudoDimensions` isn't really a place where we can store values, it is just a placeholder for where the measures should be placed instead of dimension

cbt1: Sounds great 👍

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
enell: No reason. I'll change

src/pivot-table/components/cells/ColumnAdjusterWrapper.tsx
enell: Fixed

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
enell: Yes. I need the fitToContent for dimensions before calculating the fitToContent for measures

src/pivot-table/hooks/use-data-model.ts
enell: Yes

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
haxxmaxx: Yeah it's just that it's not necessarily a fitToContent value you get back from this function. If a measure has specified a width in e.g. pixel and that is the widest one, that is the value this function returns. so it is basically `maxMeasureColumnWidth` or something

src/pivot-table/data/extract-headers.ts
enell: Yep

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
enell: Got it. I'll change the name

enell: I've changed the name

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
```diff
@@ -45,47 +77,35 @@ export default function useColumnWidthLeft({ layoutService, tableRect, headersDa
...
             }
+          }
 
-            return Math.max(maxWidth, cellWidth);
-          }, width);
+          if (columnWidth && columnWidth.type !== ColumnWidthType.Auto) {
+            width = getColumnWidthValue(tableRect.width, columnWidth, fitToContentWidth);
+          } else {
+            width = fitToContentWidth;
+          }
```
haxxmaxx, line 108: I think you don't need the else case. getColumnWidthValue should basically have the same logic, when columnWidth is undefined or type is auto, it will use return fitToContentWidth

haxxmaxx (commented):
Couple more things before I approve 🙃

src/pivot-table/data/extract-headers.ts
```diff
@@ -112,13 +120,27 @@ const extractHeaders = (
...
+
+        if (cell.dimensionInfoIndex === PSEUDO_DIMENSION_INDEX) {
+          cell.columnWidthLocation = ColumnWidthLocation.Measures;
+        } else if (!cell.isLeftDimension) {
+          if (visibleLeftDimensionInfos.at(-1) === PSEUDO_DIMENSION_INDEX) {
+            // pseudo dimension is last in left dimensions
+            cell.columnWidthLocation = ColumnWidthLocation.Measures;
+          } else if (layoutService.hasPseudoDimOnLeft) {
+            // pseudo dimension is anywhere els in left dimensions
+            cell.columnWidthLocation = ColumnWidthLocation.Pivot;
```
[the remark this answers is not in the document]
enell: Done

src/pivot-table/data/extract-headers.ts
```diff
@@ -112,13 +120,27 @@ const extractHeaders = (
...
         cell.colIdx = cell.colIdx || colIdx;
+
+        if (cell.dimensionInfoIndex === PSEUDO_DIMENSION_INDEX) {
+          cell.columnWidthLocation = ColumnWidthLocation.Measures;
+        } else if (!cell.isLeftDimension) {
+          if (visibleLeftDimensionInfos.at(-1) === PSEUDO_DIMENSION_INDEX) {
+            // pseudo dimension is last in left dimensions
+            cell.columnWidthLocation = ColumnWidthLocation.Measures;
+          } else if (layoutService.hasPseudoDimOnLeft) {
+            // pseudo dimension is anywhere els in left dimensions
```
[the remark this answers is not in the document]
enell: Fixed

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
enell: You are correct

src/pivot-table/hooks/use-column-width/use-column-width-left.ts
```diff
@@ -45,47 +77,35 @@ export default function useColumnWidthLeft({ layoutService, tableRect, headersDa
...
-
-            const lastRowLastColumn = rowIdx === headersData.size.y - 1 && collIdx === headersData.size.x - 1;
-            let cellWidth = 0;
-
-            if (header.id === PSEUDO_DIMENSION_KEY && header.isLeftDimension) {
-              // Use the max width of all measures
-              cellWidth = maxMeasureCellWidth;
+          let columnWidth: ColumnWidth | undefined;
+          if (lastRowHeader.columnWidthLocation === ColumnWidthLocation.Pivot) {
+            columnWidth = layoutService.layout.qHyperCube.topHeadersColumnWidth;
```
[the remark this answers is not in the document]
enell: Done

src/pivot-table/hooks/__tests__/use-data-model.test.ts
```diff
@@ -181,13 +181,15 @@ describe("useDataModel", () => {
 
     test("should call applyPatches with qPath for dimension", () => {
       const { applyColumnWidth } = renderer();
+      cellInfo.columnWidthLocation = ColumnWidthLocation.Dimension;
```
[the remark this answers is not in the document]
haxxmaxx: that makes sense. I'll take that into account in the future. I've written test with a minimal amount of "overrides" in each test case, so the base case is set in the beforeEach to not do it every time. But you are right, that is probably harder to read

haxxmaxx (approved):

PR #480, opened 2023-11-27
fix: empty cells after click to scroll

cbt1:
Fixes an issue where data would not be fetched if the user clicked on the scrollbar to jump to a position. So all cells would be empty.

This was because, for some charts, not sure why only some where affected, the `onScroll` event fired after the `onItemsRendered` event that tried to fetch data.

Changes:
- Setting the vertical and horizontal scroll direction is now done in `useScroll` instead of `useScrollDirection` to guarantee that the scroll direction is set before data is fetched.
- Deleted `useScrollDirection`.


enell (approved):

enell:
LGTM

PR #481, opened 2023-11-27
fix: scrollbar overlapping cells

cbt1:
The width of the vertical scrollbar was only set once, when the chart was first rendered. If a scrollbar was later added because some user action, ex: re-sizing the chart, the width of the scrollbar was not taken into when calculating sizes.

Known actions that can triggered a scrollbar to appear or disappear:
- Change theme
- Re-size chart
- Expand/Collapse a cell


src/pivot-table/components/grids/LeftGrid.tsx
```diff
@@ -71,11 +71,9 @@ const LeftGrid = ({
...
   } = useStyleContext();
 
-  useOnPropsChange(() => {
-    if (leftGridRef.current) {
-      leftGridRef.current.forEach((list) => list?.resetAfterIndex(0, false));
-    }
-  }, [dataModel, width, height, leftDimensionData, leftGridRef, contentCellHeight]);
+  useResetListCache(leftGridRef, dataModel, leftDimensionData);
+
+  useResetListCacheAndRerender(leftGridRef, width, height, contentCellHeight);
```
cbt1, line 76: When the width/height changes, the internal cache of react-windows is stale. This resets that cache and forced the grid to rerender.

It was actually already done in the `DataGrid` which could cause the "fix" in this PR put the `TopGrid` and `DataGrid` out of sync. As `TopGrid` would render with stale width values.

src/pivot-table/components/grids/TopGrid.tsx
```diff
@@ -62,11 +62,9 @@ const TopGrid = ({
...
   };
 
-  useOnPropsChange(() => {
-    if (topGridRef.current) {
-      topGridRef.current.forEach((list) => list?.resetAfterIndex(0, false));
-    }
-  }, [dataModel, width, height, topDimensionData, topGridRef, headerCellHeight]);
+  useResetListCache(topGridRef, dataModel, topDimensionData);
+
+  useResetListCacheAndRerender(topGridRef, width, height, headerCellHeight);
```
haxxmaxx, line 67: Do we need to run both? so resetting twice?

cbt1: I'm looking into just running `useResetListCache` for both data and width/height changes. But both scenarios needs to reset the cache

cbt1: I've removed `useResetListCacheAndRerender`. I'm thinking that `useResetListCache` will run before the react window lists in `LeftGrid` and `TopGrid` renders. So the cache should be reset by then.

cbt1: And some more testing have proven that `useResetListCacheAndRerender` is actually needed. So I will re-add that.

Expand/Collapsing a cell that triggered the vertical scrollbar to appear causes issues when `useResetListCacheAndRerender` is not there

haxxmaxx (approved):

PR #483, opened 2023-11-28
Crl/opaque adjuster

haxxmaxx:


src/locale/__tests__/translations.test.ts
```diff
@@ -15,12 +15,12 @@ describe("translations", () => {
       };
     });
 
-    it("Should call add for every key", () => {
+    test("Should call add for every key", () => {
```
haxxmaxx, line 18: I forgot if we were using it or test, so did a global search and these were the only ones using it, so changed them

src/pivot-table/components/cells/utils/get-adjuster-color.ts
```diff
@@ -0,0 +1,28 @@
...
+ * The background is assumed to be opaque.
+ * If the background is transparent (which is the default color) it is assumed to be white
+ */
+export default function getAdjusterColor(styleService: StyleService, isHeader: boolean) {
+  const background = isHeader ? styleService.header.background : styleService.dimensionValues.background;
+  const opaqueBackground = background === Colors.Transparent ? COLORING.WHITE : background;
+  const borderColor = color(styleService.grid.border)?.rgb();
+  const backgroundColor = color(opaqueBackground)?.rgb();
+
+  if (backgroundColor && borderColor && borderColor.opacity < 1) {
```
cbt1, line 19: Think you can do

```suggestion
  if (backgroundColor && borderColor?.opacity < 1) {
```

src/pivot-table/components/cells/utils/get-adjuster-color.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Colors } from "../../shared-styles";
+
+/**
+ * Converts border color to opaque color,
+ * by alpha blending the border color with the background the header or top grid background.
+ * The background is assumed to be opaque.
+ * If the background is transparent (which is the default color) it is assumed to be white
+ */
+export default function getAdjusterColor(styleService: StyleService, isHeader: boolean) {
+  const background = isHeader ? styleService.header.background : styleService.dimensionValues.background;
```
cbt1, line 14: Note that this wont work when the `isHeader` is false and the top dimension have background set via color by expression.

I think it's fine that we cannot handle that, the question is if it's better to do nothing in that case or use `styleService.dimensionValues.background` anyway?

src/pivot-table/components/cells/utils/get-adjuster-color.ts
```diff
@@ -0,0 +1,28 @@
...
+  const opaqueBackground = background === Colors.Transparent ? COLORING.WHITE : background;
+  const borderColor = color(styleService.grid.border)?.rgb();
+  const backgroundColor = color(opaqueBackground)?.rgb();
+
+  if (backgroundColor && borderColor && borderColor.opacity < 1) {
+    const r = borderColor.opacity * borderColor.r + backgroundColor.r * (1 - borderColor.opacity);
```
cbt1, line 20: You could store `(1 - borderColor.opacity)` in a variable. As it's used in 3 places

cbt1 (approved):
LTGM, just minor comments and something to think about regarding color by expression

src/pivot-table/components/cells/utils/get-adjuster-color.ts
haxxmaxx: I think I got a lint error, will check

src/pivot-table/components/cells/utils/get-adjuster-color.ts
haxxmaxx: I think I can fix that pretty easily, will get on it.

src/pivot-table/components/cells/utils/get-adjuster-color.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Colors } from "../../shared-styles";
+
+/**
+ * Converts border color to opaque color,
+ * by alpha blending the border color with the background the header or top grid background.
+ * The background is assumed to be opaque.
+ * If the background is transparent (which is the default color) it is assumed to be white
+ */
+export default function getAdjusterColor(styleService: StyleService, cellInfo: AdjusterCellInfo) {
+  const topGridBackground = cellInfo.expressionColor?.background || styleService.dimensionValues.background;
```
cbt1, line 14:
```suggestion
  const topGridBackground = cellInfo.expressionColor?.background ?? styleService.dimensionValues.background;
```

src/pivot-table/components/cells/utils/get-adjuster-color.ts
haxxmaxx: then borderColor can be undefined, we don't want that

src/types/types.ts
```diff
@@ -85,6 +85,7 @@ export interface AdjusterCellInfo {
   canBeResized: boolean;
   x?: number;
   colIdx?: number;
+  expressionColor?: ExpressionColor;
```
cbt1, line 88: Should you not add this to the actual AdjusterCellInfo instances as well?

src/pivot-table/components/cells/utils/get-adjuster-color.ts
haxxmaxx: Can it be empty string or?

src/pivot-table/components/cells/utils/get-adjuster-color.ts
cbt1: But if `borderColor` is undefined in `borderColor?.opacity < 1`. It should always faild the less then check?

src/pivot-table/components/cells/utils/get-adjuster-color.ts
haxxmaxx: it doesn't hurt in any case

src/types/types.ts
haxxmaxx: Not needed, in the dimension value case, we use `Cell`, so it is already included. I only specify which types the function needs.

src/pivot-table/components/cells/utils/get-adjuster-color.ts
haxxmaxx: fixed

PR #484, opened 2023-11-29
fix: data fetch on resize

cbt1:
Fixes an issue where data would not be fetch when re-sizing the PVT.

Two reason this happend:
- Both vertical and horizontal scroll direction can be reset to `None` when a PVT is re-sized.
- `isMissingColumnData` logic did not work properly as extracting "rows" from an array using `slice()` does not work if the end index is larger then the size of the array. Ex: `[0, 1, 2].slice(2, 4)` => `[2]`, but for the logic to work it should have been an array like this `[2, undefined]`

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/f6522e6a-3c12-40e4-ae58-22a6e792de1e

### DIFF
--- a/src/pivot-table/hooks/use-items-rendered-handler/utils/__tests__/fetch-pages.test.ts
+++ b/src/pivot-table/hooks/use-items-rendered-handler/utils/__tests__/fetch-pages.test.ts
@@ -46,7 +46,10 @@ describe("fetchPages", () => {
     jest.resetAllMocks();
   });
 
-  test("should not fetch pages if both scroll directions are None", async () => {
+  test("should fetch pages if both scroll directions are None", async () => {
+    horizontalScrollDirection.current = ScrollDirection.None;
+    verticalScrollDirection.current = ScrollDirection.None;
+    mockedGetRowPages.mockReturnValue([{ qLeft: 0, qTop: 0, qWidth: 1, qHeight: 1 }]);
     await fetchPages(
       dataModel,
       layoutService,
@@ -57,7 +60,7 @@ describe("fetchPages", () => {
       horizontalScrollDirection,
     );
 
-    expect(dataModel.fetchPages).toHaveBeenCalledWith([]);
+    expect(dataModel.fetchPages).toHaveBeenCalledWith([{ qLeft: 0, qTop: 0, qWidth: 1, qHeight: 1 }]);
   });
 
   test("should fetch both row and column pages", async () => {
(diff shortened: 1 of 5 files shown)


src/pivot-table/hooks/use-items-rendered-handler/utils/fetch-pages.ts
```diff
@@ -60,6 +60,14 @@ export const fetchPages = async (
     );
   }
 
+  // A chart have been re-size which have triggered a onScroll event and reset both scroll directions
+  if (
+    verticalScrollDirection.current === ScrollDirection.None &&
+    horizontalScrollDirection.current === ScrollDirection.None
```
haxxmaxx, line 66: Is this only true when the chart has been resized. Do we always fetch pages when one of the scroll directions are not `None`

src/pivot-table/hooks/use-items-rendered-handler/utils/get-row-pages.ts
```diff
@@ -36,8 +36,8 @@ const getRowPages = (
       const page = {
         qLeft,
         qTop: Math.max(0, top + pageInfo.page * pageInfo.rowsPerPage),
-        qHeight: 1,
         qWidth,
+        qHeight: 1,
```
haxxmaxx, line 40: any reason for this change?

src/pivot-table/hooks/use-items-rendered-handler/utils/__tests__/fetch-pages.test.ts
```diff
@@ -46,7 +46,10 @@ describe("fetchPages", () => {
     jest.resetAllMocks();
   });
 
-  test("should not fetch pages if both scroll directions are None", async () => {
+  test("should fetch pages if both scroll directions are None", async () => {
```
haxxmaxx, line 49: I assume this test didn't catche the issue before?

haxxmaxx (approved):

src/pivot-table/hooks/use-items-rendered-handler/utils/get-row-pages.ts
cbt1: Yes hovering the return values of `getRowPages` in VCS the output type would be `{ qLeft, qTop, qHeight, qWidth }`. It's now changed to `{ qLeft, qTop, qWidth, qHeight }` which is the "correct" property order.

src/pivot-table/hooks/use-items-rendered-handler/utils/fetch-pages.ts
cbt1: It will be `None` for both in the default state. So when the chart is first rendered. But should end with all data already being fetched. So no additonal fetch there.

PR #485, opened 2023-11-29
fix: column width for empty row

cbt1:
Fixes an issue where the column width was not correctly calculated for the "leaf row" when the last row is an empty row.

Before

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/b3f42ff8-17fd-4d3f-b64d-4baead79bb95

After

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/fcbd9a56-5897-4349-8b5b-560a2c7f40c7

### DIFF
--- a/src/pivot-table/components/helpers/get-item-size-handler.ts
+++ b/src/pivot-table/components/helpers/get-item-size-handler.ts
@@ -42,7 +42,7 @@ export const getColumnWidthHandler =
     }
 
     // all rows except bottom one
-    if (cell?.leafCount > 0) {
+    if (!isLastRow && cell !== undefined) {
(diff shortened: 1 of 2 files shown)


src/pivot-table/components/helpers/get-item-size-handler.ts
```diff
@@ -42,7 +42,7 @@ export const getColumnWidthHandler =
     }
 
     // all rows except bottom one
-    if (cell?.leafCount > 0) {
+    if (!isLastRow && cell !== undefined) {
```
cbt1, line 45: "Leaf row", the row before the empty row never evaluate this to true since all cells in a "leaf row" have 0 as `leafCount`.

enell (approved):

PR #486, opened 2023-11-30
fix: empty cells after expanding or collapsing cell

cbt1:
Fixes an issue where expanding or collapsing a cell could result in too few rows or columns being fetched and empty cells being rendered.

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -1,7 +1,5 @@
 export const Q_PATH = "/qHyperCubeDef";
 
-export const DEFAULT_PAGE_SIZE = 50;
-
 export const TOTALS_CELL = -1;
 
 export const TOTAL_MODE_ON = "TOTAL_EXPR";
--- a/src/hooks/use-load-data-pages.ts
+++ b/src/hooks/use-load-data-pages.ts
@@ -76,12 +76,57 @@ export const getFetchArea = (
   let qLeft = 0;
   let qTop = pageStartIndex;
 
+  /**
+   * Things that can happen
+   * qLastExpandedPos equals the position at which was expanded or collapsed. No data was added or removed before the position.
+   * qLastExpandedPos does not equal the position as which it was expanded/collapsed. Data was added or removed before the position.
+   *
+   * Data before the position is removed: viewService.gridColumnStartInde and viewService.gridRowStartIndex are now STALE.
+   * As the position might no longer exist. Can that be checked?
+   *
+   * Data before the position is added: viewService.gridColumnStartInde and viewService.gridRowStartIndex are still valid.
+   * But might not be where the cell was expanded.
+   */
+
   // qLastExpandedPos only exist in the layout if a new layout was received because a node was expanded or collapsed
   if (qLastExpandedPos) {
+    if (viewService.lastExpandedOrCollapsed.grid === "left") {
+      const lastRow = viewService.gridRowStartIndex + viewService.gridHeight;
+
+      if (qLastExpandedPos.qy === viewService.lastExpandedOrCollapsed.rowIndex) {
+        // No data was added before or after the expand/collapse position
+        console.log("%c No data was added before or after the expand/collapse position", "color: lime");
+        qTop = pageStartIndex + viewService.gridRowStartIndex;
+      } else if (qLastExpandedPos.qy > viewService.lastExpandedOrCollapsed.rowIndex) {
+        // Data was added before the expanded position
+        console.log("%c Data was added before the expanded position", "color: salmon");
+        qTop = pageStartIndex + viewService.gridRowStartIndex;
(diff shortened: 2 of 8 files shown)


src/constants/index.ts
```diff
@@ -1,7 +1,5 @@
 export const Q_PATH = "/qHyperCubeDef";
 
-export const DEFAULT_PAGE_SIZE = 50;
```
cbt1, line 3: There is no default page size any more. It's all resolved from the height and width of the chart rect.

src/hooks/use-load-data-pages.ts
```diff
@@ -74,21 +76,82 @@ export const getFetchArea = (
...
+     * - qLastExpandedPos[qx | qy] is less than the position where the cell was collapsed.
+     *   - Rows or columns was removed before collapsed cell.
+     *   - The row or column index where the cell was collapsed NO LONGER exist in the data.
+     *   - The last row or column index that was visible to the user when the cell was collapsed NO LONGER exists in the data
+     *   - Start indexes in ViewService are INVALID.
+     *   - After collapsing, cell will not be in view.
+     *
+     */
+
+    qLeft = viewService.gridColumnStartIndex;
```
cbt1, line 133: The next 10-15 lines is the actual fix

src/hooks/use-load-data-pages.ts
```diff
@@ -55,17 +55,19 @@ export const isMissingLayoutData = (
...
   pageInfo: PageInfo,
   maxNumberOfVisibleRows: number,
   maxNumberOfVisibleColumns: number,
 ) => {
-  const pageStartIndex = pageInfo.page * pageInfo.rowsPerPage;
-  const pageEndIndex = pageStartIndex + pageInfo.rowsPerPage;
+  const pageRowStartIndex = pageInfo.page * pageInfo.rowsPerPage;
+  // Do not fetch data beyond this value. Either because there is no more data in the layout or the current page ends.
+  const rowEndIndex = Math.min(pageRowStartIndex + pageInfo.rowsPerPage, qSize.qcy);
+  const columnEndIndex = Math.min(MAX_COLUMN_COUNT, qSize.qcx);
```
a-m-dev, line 68: let's imagine a scenario where we have 50k cols in dataset
`MAX_COLUMN_COUNT` is 15k and qcx shows us the above 50k
does this means when we fetch new batch of data, we will always fetch batches in 15k size in width?

probably I'm missing something but as far as i checked with line 153 in this module, seems like we just subtracting what ever current `qLeft` is from  `columnEndIndex`, not sure if `maxNumberOfVisibleColumns` will cover that scenario? 

In the other hand, seems like `maxNumberOfVisibleColumns` gives us some bigger number than we expect, for example, there is like 10 cells visible in dataGrid below, but we get 42 cells as `maxNumberOfVisibleColumns`: 

Screenshot 2023-12-04 at 11 16 49

a bit of digging into this  i notices the initial width of rect.width:

Screenshot 2023-12-04 at 11 18 14

and here is the calculation from `handle-max-engine-size`:

Screenshot 2023-12-04 at 11 20 20
 
seems like it considers the cell min pixel values and this means for every fetch -> we will endup fetching some cells more than expected?
I mean, AFAIunderstood the code, from first part of my comment -> we are rely on `maxNumberOfVisibleColumns` when it comes to fetching a new batch of data (always, but consider the large col scenario like 50k for now), and meanwhile this way, `maxNumberOfVisibleColumns` calculated based on min cell pixels size which returns more cell counts than what user can see (like without subtracting column sizes from left grid, and not the actual size of rendered cell in data grid -> because visually what users can see is ~10 to 12 cells)

a-m-dev: Yes, absolutely!
the thing that i was thinking was, when we taking the safe guess amount for cells, we dont subtract the left grid width from rect.width, but as you said at that point in render flow, we dont have access to left grid element to actually measure its width and subtract it fomr react.width, so it looks fine if we consider to fetch more cells than the real actual width is. 

one last concern could be this below scenario:

Screenshot 2023-12-04 at 11 56 54

where we could have a large data cells (for some reason), and of course large value for react.width (4k, 8k size), 
in that case the cells we will fetch is going to be 8194/30 =  ~273 ish, but user may see much less cells than that. 

That is like more unlikely edge case, but not sure if it's worth the effort + that much important to take care of.

a-m-dev (approved):
LGTM 🚀

src/hooks/use-load-data-pages.ts
cbt1: Yes there are scenarios where we end up fetching way more then needed. On an 8k resolution we will fetch ~37 000 cells, but many of them are likely not going to be visisble.

It would be nice to have some metrics on the number of rows/columns the user actually have in their data, so that we can fine-tune our solution around that.

PR #489, opened 2023-12-04
fix: default colors

cbt1:
We can no longer use colors with transparency and should instead use HEX colors.

### DIFF
--- a/src/pivot-table/components/cells/utils/__tests__/get-adjuster-color.test.ts
+++ b/src/pivot-table/components/cells/utils/__tests__/get-adjuster-color.test.ts
@@ -15,7 +15,7 @@ describe("get-adjuster-color", () => {
         background: Colors.Transparent,
       },
       grid: {
-        border: Colors.Black15,
+        border: "rgba(0, 0, 0, 0.15)",
       },
     } as StyleService;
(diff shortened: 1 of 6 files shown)


src/pivot-table/components/cells/utils/__tests__/get-adjuster-color.test.ts
```diff
@@ -15,7 +15,7 @@ describe("get-adjuster-color", () => {
         background: Colors.Transparent,
       },
       grid: {
-        border: Colors.Black15,
+        border: "rgba(0, 0, 0, 0.15)",
```
a-m-dev, line 18: I think `Colors.DividerLight` here? 
because it's already been replaced in line 25 in related test case

a-m-dev (approved):
a minor comment but LGTM

src/pivot-table/components/cells/utils/__tests__/get-adjuster-color.test.ts
cbt1: No I could not change that. The tests expected the border color to be transparent. So had to keep that value here. The scenario/s are still supported, just that the default colors are not the same any more.

PR #492, opened 2023-12-04
feat: chart background color and image

cbt1:
Adds support for setting chart background color and background image. Both via property panel and directly using properties.

Added to Background color and Background image section to property panel:

Skärmavbild 2023-12-04 kl  19 00 36

### DIFF
--- a/src/qae/initial-properties.js
+++ b/src/qae/initial-properties.js
@@ -162,6 +162,8 @@ const properties = {
  * @type object
  * @property {GeneralStylingKey} key
  * @property {TitleStyling} title
+ * @property {BackgroundColor | BackgroundColorExpression} [bgColor]
+ * @property {BackgroundImage} [bgImage]
  * @example
  * {
  *  key: "general",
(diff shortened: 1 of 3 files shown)


src/qae/initial-properties.js
```diff
@@ -172,10 +174,63 @@ const properties = {
...
+ * @name BackgroundColor
+ * @type object
+ * @property {PaletteColor} color - Background color palette
+ */
+
+/**
+ * Chart background color by expression
+ * @name BackgroundColorExpression
+ * @type object
+ * @property {boolean} useExpression - Boolean to indicate if color by expression should be used
```
cbt1, line 212: Noted that setting background expression via the property panel will use `useColorExpression` while nebula only consumes `useExpression`. Most likely a bug in Sense or in Nebula.

cbt1: Fix in nebula https://github.com/qlik-oss/nebula.js/pull/1426

a-m-dev (approved):
Just a dummy question, but do we need to do something when a chart has bg color or bg image? will color or bg image going to be handled by card container that holds the chart in sense client? am i right?

cbt1:
> Just a dummy question, but do we need to do something when a chart has bg color or bg image? will color or bg image going to be handled by card container that holds the chart in sense client? am i right?

Not dumb at all. The backgrounds are handled by the context they are running in. So when PVT is created in Sense, background is handled by Sense. When PVT is created in a mashup using nebula directy, background is handled by nebula.

So the PVT does not need to do anything else then enabled the option in the property panel.

PR #493, opened 2023-12-05
fix: font family dropdown

cbt1:
The dropdown for font family did not match the general styling drop down. This PR fixes:
- Get default fonts from the same source as the general styling drop down (`getAvailableFonts`).
- Apply the font family on the actual item on the drop down to get a preview of it.

Before:
Skärmavbild 2023-12-05 kl  10 51 49

After:
Skärmavbild 2023-12-05 kl  10 50 24

### DIFF
--- a/src/pivot-table/constants.ts
+++ b/src/pivot-table/constants.ts
@@ -1,5 +1,5 @@
 export const DEFAULT_FONT_SIZE = "12px";
-export const DEFAULT_FONT_FAMILY = "Source Sans Pro, Arial, sans-serif";
+export const DEFAULT_FONT_FAMILY = "Source Sans Pro, sans-serif";
 export const DEFAULT_CELL_HEIGHT = 24;
 export const DEFAULT_HEADER_CELL_HEIGHT = 32;
 export const DEFAULT_LINE_CLAMP = 1;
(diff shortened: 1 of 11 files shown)


src/pivot-table/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const DEFAULT_FONT_SIZE = "12px";
-export const DEFAULT_FONT_FAMILY = "Source Sans Pro, Arial, sans-serif";
+export const DEFAULT_FONT_FAMILY = "Source Sans Pro, sans-serif";
```
a-m-dev, line 2: why removing Arial?

a-m-dev: is it because of this: https://github.com/qlik-trial/qlik-chart-modules/blob/d6c824406114ced8dd7b6f113e1b1f773f9451e6/src/resources/fonts/getAvailableFonts.js#L5C18-L5C18?

a-m-dev (approved):
🚀

src/pivot-table/constants.ts
cbt1: That and the Sense themes use `"Source Sans Pro, sans-serif"`

PR #495, opened 2023-12-11
fix: use correct text measurement for measure labels

haxxmaxx:
- the measurement of measure labels was using the wrong styling. It is a bit complicated, since the `measureLabels` inherits font family and size from `dimensionVaules`, but has its own `fontStyle` and `fontWeight`
- added some logic to get the correct args, I think the hook then looks cleaner
- also the deps for `setLeftGridWidthInfo` was wrong

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -245,10 +245,10 @@ export type BasicCellStyling = {
   background: string;
 };
 
-export interface CellStyling extends BasicCellStyling {
+export type CellStyling = BasicCellStyling & {
   fontSize: string;
   fontFamily: string;
-}
+};
 
 export type StylingPanelOptions = {
   header: CellStyling;


src/types/types.ts
```diff
@@ -237,23 +237,26 @@ export interface PageInfo {
...
-  fontFamily: string;
+export type BasicCellStyling = {
   fontWeight: string | undefined; // "600" | "normal" from Styling panel, but from Theme it can be any supported font-weight value
   fontStyle: string; // "italic" | "normal" from Styling panel, but from Theme it can be any supported font-weight value
   textDecoration: string; // "underline" | "none" from Styling panel, but from Theme it can be any supported font-weight value
   color: string;
   background: string;
 };
 
+export interface CellStyling extends BasicCellStyling {
```
cbt1, line 248: I've used `type` over `interface` to get a better tooltip in VSC. I think you could do the same here

cbt1 (approved):

cbt1 (approved):

PR #496, opened 2023-12-12
fix: none color

cbt1:
Fixes an issue where setting the color for "border" and "divider" via the Styling panel to "none", would cause the borders to render with `currentcolor` producing some unexpected results.

Background color and font color appears to be working with `none` as color but this PR fixes those scenarios as well. As `none` is not a valid css color value.

Before:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/246a71e9-c01f-4a65-a017-d0671c19fe6c

After:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/c8260424-43c8-4153-b4f7-9fb64d24746a

### DIFF
--- a/src/services/style-service.ts
+++ b/src/services/style-service.ts
@@ -66,8 +66,19 @@ enum ThemeAttribute {
 
 const resolveFontSize = (fontSize: string | undefined) => (fontSize ? `${parseInt(fontSize, 10)}px` : undefined);
 
-const resolveColor = (theme: ExtendedTheme, color: PaletteColor | undefined) =>
-  color ? theme.getColorPickerColor(color) : undefined;
+export const resolveColor = (theme: ExtendedTheme, color: PaletteColor | undefined) => {
+  if (color) {
+    const resolvedColor = theme.getColorPickerColor(color);
+    // Handle when color is set to "none" via the color picker
+    if (resolvedColor === "none") {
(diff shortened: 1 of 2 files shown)


src/services/style-service.ts
```diff
@@ -66,8 +66,19 @@ enum ThemeAttribute {
...
 const resolveFontSize = (fontSize: string | undefined) => (fontSize ? `${parseInt(fontSize, 10)}px` : undefined);
 
-const resolveColor = (theme: ExtendedTheme, color: PaletteColor | undefined) =>
-  color ? theme.getColorPickerColor(color) : undefined;
+export const resolveColor = (theme: ExtendedTheme, color: PaletteColor | undefined) => {
+  if (color) {
+    const resolvedColor = theme.getColorPickerColor(color);
+    // Handle when color is set to "none" via the color picker
+    if (resolvedColor === "none") {
+      return Colors.Transparent;
```
a-m-dev, line 74: Just a question out of curiosity, why not returning undefined here? is it any different than returning transparent?

cbt1: Color is resolved in a specific order like this `resolveColorFromStylingPanel ?? colorFromTheme ?? fallbackColor`. So returning `undefined` would result in it using theme color instead.

PR #498, opened 2023-12-13
fix: cell key

cbt1:
Attempts to generate proper ID:s for all cells. Previously it was either done using col/row index for the DataGrid and using `qElemNo` for Left/Top-Grid. This PR follows the advice given at [react-window itemKey docs](https://react-window.vercel.app/#/api/FixedSizeGrid) and [React docs](https://react.dev/learn/rendering-lists#why-does-react-need-keys).

It also a real fix for #497 as the root issue was that the cell first generated at x/y in the DataGrid was replaced with a new cell, because of a new layout after expansion, that caused a change in shorthand style properties to non-shorthand style properties.


src/pivot-table/data/helpers/get-random-uuid.ts
```diff
@@ -0,0 +1,5 @@
+import { v4 as uuidv4 } from "uuid";
+
+const getRandomUUID = () => crypto?.randomUUID?.() ?? uuidv4();
```
a-m-dev, line 3: crypto seems like having a good support for [most browsers](https://developer.mozilla.org/en-US/docs/Web/API/Crypto#browser_compatibility), why should we still have uuidv4 as fallback solution?

a-m-dev (approved):

src/pivot-table/data/helpers/get-random-uuid.ts
cbt1: `randomUUID` is not available for all browsers in non-secure contexts https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

PR #499, opened 2023-12-14
fix: highlight selected values

cbt1:
A selection in a field is done on unique values, those unique values can be presented in the PVT as multiple different values (same value but different parent nodes). All selected values should be highlighted, not only the cell that was selected, but all other representations of it as well.

The key to identify a unique value in a field is `qElemNo`, this is now used to track selected cells.

Before:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/c544d857-3d6e-407a-8f24-fc86b4c3fdd1

After:

https://github.com/qlik-oss/sn-pivot-table/assets/16608020/aba994d4-35b9-4c40-a078-c0137da14299

### DIFF
--- a/src/pivot-table/hooks/use-selections-model.ts
+++ b/src/pivot-table/hooks/use-selections-model.ts
@@ -14,22 +14,16 @@ export interface SelectionModel {
   isLocked: SelectionCellLookup;
 }
 
-export interface SelectedPivotCell {
-  qType: string;
-  qRow: number;
-  qCol: number;
-}
-
 type SelectedField = {
   selectionCellType: NxSelectionCellType;
   coord: number;
 };
 
-const getNextState = (cell: Cell, selectedPivotCells: Set<Cell>, selectedField: SelectedField | null) => {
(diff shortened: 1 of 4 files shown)


src/pivot-table/hooks/use-selections-model.ts
```diff
@@ -140,7 +136,10 @@ export default function useSelectionsModel(
     [selections, isLocked, selectedPivotCells, selectedField],
   );
 
-  const isSelected = useCallback((cell: Cell) => selectedPivotCells.has(cell), [selectedPivotCells]);
+  const isSelected = useCallback(
+    (cell: Cell) => !isLocked(cell) && selectedPivotCells.has(cell.ref.qElemNo),
```
haxxmaxx, line 140: why not check isActive here, so we don't have do to that in dimensionValue?

cbt1: Genius! Will fix

haxxmaxx (approved):

PR #500, opened 2023-12-14
perf(useColumnWidth): remove overuse of memoization and make no unnesceary updates of return values

haxxmaxx:
The goal with this PR is that **nothing returned from the `useColumnWidth` hook should change (new value or instance), when we need the pivot table to be performant**. By performant I mean doing interactions with the chart, not editing it (then a full rerender is totally OK). For that I include
- selections
- scrolling
- sorting
- column resize

The resize will trigger a update in any case, for obvious reasons. Sorting fetches new data, so that one is hard to prevent. But for scrolling and selections, there should be no new values or instances returned. So now the properties returned from this hook, wont trigger any rerenders for the child components of `PivotTable`

What I've changed:
- only passing the `tableWidth`, the height is not relevant
- removing useMemo for values that are cheap to calculate, and since they are non-objects, they don't trigger other hooks
- moved destructuring of `layoutService` inside `calculateLeftGridWidthInfo`, to simplify the dependency array
- being more specific in the dep array for `getHeaderCellsIconsVisibilityStatus`, that one was actually updating every time it ran

### DIFF
--- a/src/pivot-table/components/PivotTable.tsx
+++ b/src/pivot-table/components/PivotTable.tsx
@@ -101,7 +101,7 @@ export const StickyPivotTable = ({
     overrideLeftGridWidth,
   } = useColumnWidth({
     layoutService,
-    tableRect,
+    tableWidth: tableRect.width,
     headersData,
     visibleTopDimensionInfo,
     verticalScrollbarWidth,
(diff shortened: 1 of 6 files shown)


cbt1 (approved):
LGTM

Did you notice that it reduced the number of renders?

haxxmaxx:
> Did you notice that it reduced the number of renders?

I actually didn't check. I could do that.
I don't think it will be any difference, since there are other things triggering an update anyway. The main thing is that I could remove some memoized stuff and still get the same result from the hook

PR #503, opened 2023-12-21
perf: min max buffer

cbt1:
This PR attemps to reduce the number of re-renders that happens because data is fetched.

- Moved the area logic from `fetch-pages.ts` into `get-column-pages.ts` and `get-row-pages.ts`
- Added max buffer logic in `get-column-pages.ts` and `get-row-pages.ts`

Before:
Every time a new row or column that had cells without data was scrolled into "view", data was fetched for that row/column (+ min buffer). This means that scrolling 100 rows/column would at most trigger 100 re-renders.

After:
Every time a new row or column that has cells without data is scrolled into "view", data is fetched for that row/column (+ min buffer + max buffer). This means that scrolling 100 rows/column will at most only trigger 4 re-renders.

Skärmavbild 2023-12-21 kl  09 49 41


a-m-dev:
I'm actually having a bit trouble understanding the concept of this PR, but i think this part here: 

> This means that scrolling 100 rows/column would at most trigger 100 re-renders.

does not happens really, because if you scroll into a row/col which does not has the data yet, it will try to fetch from that starting point(what ever row/col you landed on, at least as far as i know about the data fetching flow). 

and here is the current behaviour on main, which shows that even scrolling into a completely empty area, does not re-render 100 or n times (n is the row counts in that empty area): 

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/2ca8d2da-993b-41c0-9d8e-db40556149bb

So even if we fetch the max buffer area, potentially we are fetching more data that expected maybe? 


Also regarding this part: 

> This means that scrolling 100 rows/column will at most only trigger 4 re-renders

I think the min buffer's purpose is exactly what we want to achieve with max buffer! why not instead, choose a bigger minBuffer? since the scenario of fetching new data will **always** happen with both min and max buffer? 
and in that case, what would be the main diff between min and max buffer in essence?

cbt1:
@a-m-dev I've been struggling to describe how and why this works as it does. So I'll give another try here for some of your concerns.


> I'm actually having a bit trouble understanding the concept of this PR, but i think this part here:
> 
> > This means that scrolling 100 rows/column would at most trigger 100 re-renders.
> 
> does not happens really, because if you scroll into a row/col which does not has the data yet, it will try to fetch from that starting point(what ever row/col you landed on, at least as far as i know about the data fetching flow).

The "at most" of what I said is important here. To simply, a user can do 2 kinds of scrolling actions that will trigger data to be fetched - 1. User scrolls slowly so that "every" row or column is triggering a data fetch. 2. User scrolls quickly (what you did) to a new position, this means that "throttled data fetching logic" will get triggered and skip some of the scroll events that could have trigger a data fetch.

So when I said 100 re-renders I meant the slow scrolling example where you actual get 100 data fetches, thus 100 re-renders. In you example the number of re-renders will likely barely improve since the "throttled data fetching logic" is already handling that fairly well.

> Also regarding this part:
> 
> > This means that scrolling 100 rows/column will at most only trigger 4 re-renders
> 
> I think the min buffer's purpose is exactly what we want to achieve with max buffer! why not instead, choose a bigger minBuffer? since the scenario of fetching new data will **always** happen with both min and max buffer? and in that case, what would be the main diff between min and max buffer in essence?

The MIN buffer and the MAX buffer have different purposes. The MIN buffer is there to improve the user experience when the user scroll slowly (see the 1. description above), it reduces the chance of rendering empty cells when no data have been fetched yet. That is why the min buffer extends the rendered area a little bit, it triggers data to be fetched for an area that is yet not visible to the user. So when they scroll that area into view, the data have already been fetched.

The MAX buffer is there to re-reduce the number of times data is fetched. For this to make some sense it important to understand when data is fetched. The flow for fetching data looks something like this: `Scroll Event -> Scroll react-window components -> react-window's onItemsRendered callback checks if any rendered cells are missing data -> Fetch data for missing rows or columns -> Re-render "all" components`. This is why scrolling slowly can trigger data to be fetched for every "empty" row or column. The MAX buffer will make it so that the `react-window's onItemsRendered callback checks if any rendered cells are missing data` step less often finds cells with missing data. Resuling in fewer re-renders.

Not sure if I can explain this any better. If there are still some uncertainties we can have a chat about it?

a-m-dev (approved):
Thanks for the extra explanation, LGTM 🚀

PR #505, opened 2023-12-21
fix: scrollbar from column adjuster overflow

cbt1:
Fixes the horizontal scrollbar that appears because of the an overflow issue with the column adjuster

Before:
Skärmavbild 2023-12-21 kl  14 38 23

After:
Skärmavbild 2023-12-21 kl  14 37 57

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "@playwright/test": "1.40.1",
     "@qlik-trial/sprout": "3.5.0",
     "@qlik/eslint-config": "0.6.2",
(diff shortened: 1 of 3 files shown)


a-m-dev:
is this the same as this ticket VIZ-2867 in jira?

https://github.com/qlik-oss/sn-pivot-table/assets/29652890/6ad63a56-10f0-43f8-b440-7864e4c2b077


because i have found a clue that it was only happening when we had stacked columns

haxxmaxx (approved):

PR #506, opened 2023-12-22
fix: null value representaiton only for measure values

cbt1:
The primary purpose of the null value representation property was to allow the user to change the measure values from `-` to `0` instead. It does not make all that much sense to have it combined with the dimension values.

I think for now we can have it only affect the measure values, if need we can add a new property for the dimension values.


a-m-dev (approved):
LGTM

cbt1:
Since this is a customer impacting functional changes, it needs to be behind a feature flag. Working on that.

a-m-dev (approved):

PR #507, opened 2023-12-22
fix: handle hidden measures

cbt1:
It's currently not possible to set a show conditions on measures via the property panel, but by either convering an old PVT to sn-pivot-table or using a mashup with calculated conditions properties it's still possible to get layout with hidden measures.

This PR fixes:
-  An issue that can occur when combining color by expression on a measure and at the same time have a hidden measure before (with lower index) the measure with color by expression.
- An issue that can occur when adjusting the column width of a measure when a hidden measure exist before the adjusted measured. The column width patch would be applied to the wrong qMeasureInfo.
- With pseudo dimension on the left side. The column width would in include hidden measures, potentially having great affect on the column width.
- There might be other issues as well but the above issues are the once that I've found.

The fix:
- Filter `qMeasureInfo` by looking for a calculated condition error code. Consume the filtered (`visibleMeasureInfo`) instead of `qMeasureInfo`.

### DIFF
--- a/src/pivot-table/components/helpers/get-item-size-handler.ts
+++ b/src/pivot-table/components/helpers/get-item-size-handler.ts
@@ -33,7 +33,7 @@ export const getColumnWidthHandler =
   ({ list, listValues, isLastRow, getRightGridColumnWidth }: ColumnWidthHandlerProps): ItemSizeHandler =>
   (colIndex: number) => {
     const cell = isLastRow ? list[colIndex] : listValues[colIndex];
-    // const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
+    // const measureInfoCount = layoutService.visibleMeasureInfo.length;
 
     // TODO: This a bit of a special case but if you are on a different page then the first.
     // Scroll down a bit and expand a node. The "first row" does not exist yet as data has only been
--- a/src/services/layout-service.ts
+++ b/src/services/layout-service.ts
@@ -3,6 +3,9 @@ import { MAX_COLUMN_COUNT, MAX_ROW_COUNT } from "../pivot-table/constants";
 import { type PivotLayout } from "../types/QIX";
 import type { LayoutService, VisibleDimensionInfo } from "../types/types";
 
+// Error code from Engine in qDimensionInfo and qMeasureInfo when calculated condition (show condition) resolves to false
+const HIDDEN_ERROR_CODE = 7005;
+
 const createLayoutService = (
   layout: PivotLayout,
   effectiveProperties: EngineAPI.IGenericObjectProperties | undefined,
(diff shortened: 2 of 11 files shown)


src/services/layout-service.ts
```diff
@@ -22,16 +25,22 @@ const createLayoutService = (
...
         return 0;
       }
 
-      return index % qMeasureInfo.length;
+      const visibleMeasureInfoIndex = index % visibleMeasureInfo.length;
+      if (getVisibleIndex) {
+        return visibleMeasureInfoIndex;
+      }
+
+      return qMeasureInfo.findIndex((info) => info === visibleMeasureInfo[visibleMeasureInfoIndex]);
```
a-m-dev, line 43: I think this part could be true for any object: 
```tsx
info === visibleMeasureInfo[visibleMeasureInfoIndex]
```

because `visibleMeasureInfo` is resulted from `.filter(...)` which will be a new set of arrays, so i think even references of those measure objects will not match together? 

probably an alternate solution could be checking the `cId` or `qLibraryId` or something else within the measure obj.

src/pivot-table/components/helpers/get-item-size-handler.ts
```diff
@@ -33,7 +33,7 @@ export const getColumnWidthHandler =
   ({ list, listValues, isLastRow, getRightGridColumnWidth }: ColumnWidthHandlerProps): ItemSizeHandler =>
   (colIndex: number) => {
     const cell = isLastRow ? list[colIndex] : listValues[colIndex];
-    // const measureInfoCount = layoutService.layout.qHyperCube.qMeasureInfo.length;
+    // const measureInfoCount = layoutService.visibleMeasureInfo.length;
```
a-m-dev, line 36: maybe remove this line? (or do we keep it there for future reference?)

src/services/layout-service.ts
cbt1: It will match since the object reference inside `qMeasureInfo[]` and `visibleMeasureInfo[]` are still the same.

It should work like this:
```js
const m0 = { visible: false };
const m1 = { visible: true };
const m2 = { visible: true };

const qMeasureInfo  = [m0, m1, m2];
const visibleMeasureInfo = qMeasureInfo.filter(info => info.visible);

return qMeasureInfo.findIndex(info => visibleMeasureInfo[1]); // Returns index 2 for "m2"
```

src/pivot-table/components/helpers/get-item-size-handler.ts
cbt1: I will keep it for now. Would need to investigate if that that will be needed down the road.

src/services/layout-service.ts
a-m-dev: if you run that code in console -> the findIndex will give you 0 matching the first object in it's search flow through `qMeasureInfo`, and that is the problem that was mentioning

cbt1: Will take a look

cbt1: It was a typo in the example. Should work with this fix:

```js
const m0 = { visible: false };
const m1 = { visible: true };
const m2 = { visible: true };

const qMeasureInfo  = [m0, m1, m2];
const visibleMeasureInfo = qMeasureInfo.filter(info => info.visible);

return qMeasureInfo.findIndex(info => info === visibleMeasureInfo[1]); // Returns index 2 for "m2"
```

a-m-dev (approved):

PR #510, opened 2024-01-10
fix: color by expression for pseudo dimensions

cbt1:
Fixes an issue where the color by expression (text and/or background) for a measure would not be properly applied if the PVT had pseudo dimensions.

If would happen if the following condition was true:
- Pseudo dimensions where not on the last row or column
- One or more measures before a measure with color by expression can not have color by expression.

#### The fix

The old logic to get measure info index via `layoutService.getMeasureInfoIndexFromCellIndex()` would only work IF the pseudo dimensions where leaf nodes in either the left or top grid. The reason is that  the number of pseudo dimension nodes might not evenly match the number of measure nodes.

To get the correct measure info index we need to look at the dimension nodes coming from `qLeft` and `qTop` in the pivot data pages. They have a `qElemNo` number which will match the index of the measure info. But this means that the `DataCell` component needs to resolve the measure info index at rendering time from the dimension cells in `TopDimensionData` and `LeftDimensionData`.

#### Before/After

In the screenshots below there are 3 measures called `=0`, `=1` and `=2` where `=2` have color by expression set to `green()`. Notice how measure cells for `=2` does not get the correct background color in the before screenshot.

Before:

Skärmavbild 2024-01-10 kl  09 16 46

After:
Skärmavbild 2024-01-10 kl  09 14 17


cbt1:
> tbh, it took me a while to understand this PR (or maybe i'm not sure if i understood it fully yet), but code and idea looks solid.
> 
> Just a side note on `getMeasureInfoIndexFromCellIndex(...)` where also you have comment on, maybe we can have a unified way of finding index info for measures in future in a hook?

Yeah maybe, I did not really wanted to restructure both `useDataModel` and `useColumnWidth` for a case that was already working for them. But in the long term the current solution might be a bit confussing.

PR #517, opened 2024-01-18
refactor: use-snapshot.ts

Donya-qlik:


src/hooks/use-snapshot.ts
```diff
@@ -12,38 +12,27 @@ interface UseSnapshotProps {
...
-      return copyOfLayout;
+      snapshotLayout.snapshotData.object.size.w = rect.width;
+      snapshotLayout.snapshotData.object.size.h = rect.height;
     }
-
-    return copyOfLayout;
+    return snapshotLayout;
   });
 
   if (layoutService.layout.snapshotData?.content) {
```
Donya-qlik, line 38: I'm not sure when this condition is used ?! 🤔

cbt1: Don't quite remember but could be when the chart is re-sized in storytelling

src/hooks/use-snapshot.ts
```diff
@@ -12,38 +12,27 @@ interface UseSnapshotProps {
 }
 
 const useSnapshot = ({ layoutService, viewService, rect, model }: UseSnapshotProps): stardust.Rect => {
-  onTakeSnapshot(async (copyOfLayout: SnapshotLayout) => {
-    if (!copyOfLayout.snapshotData) {
-      return copyOfLayout;
+  onTakeSnapshot(async (snapshotLayout: SnapshotLayout) => {
+    if (!snapshotLayout.snapshotData || !model || snapshotLayout.snapshotData.content) {
```
cbt1, line 16: Is the `content` object always there?

cbt1: Just did some quick testing and it does seem like there is no `content` the first time `onTakeSnapshot` is called

src/hooks/use-snapshot.ts
cbt1: Did some testing and cannot find a case where this logic is triggered. For now lets keep it but we might be able to remove it in the feature

src/hooks/use-snapshot.ts
cbt1: Just had another look. You are indeed correct 👍

cbt1 (approved):

PR #520, opened 2024-01-21
feat(IM-5878): consider scroll position when taking snapshot

Donya-qlik:
**What is the improvement ?**:
Pivot table snapshot should consider the scroll position so WYSIWYG as the result.

**How it is improved:**
The idea is when the snapshot is taken, `snapshotData content` provides one more property `rowPartialScrollHeight ` before the table re-renders. The considered property is basically a calculation for finding the hidden part of the table row when the scroll is positioned and add the result to the scrollTop/Left values when the snapshot is used.

Before merging this PR, all get-view-state properties will be reset to 0 to ensure they do not affect the current snapshot state. This will remain the case until further calculations are added

The attachment shows how the pv table looks like when the row partial height is applied on the snapshot.
Skärmbild 2024-01-22 004022

### DIFF
--- a/src/ext/index.ts
+++ b/src/ext/index.ts
@@ -8,7 +8,7 @@ export default function ext(env: Galaxy): Record<string, unknown> {
       // TODO Most of the code is in place to enable snapshot, however WYSIWYG is not always true. Depending on the scroll position, some cell that where visible when the screenshot was taken, is not fully visiable in the screenshot.
       // This should be fixed, possibly be figuring out how much of a cell is in view and then in the snapshot scroll that cell into view by the same amount.
       cssScaling: false,
-      snapshot: false,
+      snapshot: env.flags.isEnabled("CLIENT_IM4976_PIVOT_DOWNLOAD"),
       export: false,
       sharing: false,
       exportData: true,
--- a/src/ext/index.ts
+++ b/src/ext/index.ts
@@ -8,7 +8,7 @@ export default function ext(env: Galaxy): Record<string, unknown> {
       // TODO Most of the code is in place to enable snapshot, however WYSIWYG is not always true. Depending on the scroll position, some cell that where visible when the screenshot was taken, is not fully visiable in the screenshot.
       // This should be fixed, possibly be figuring out how much of a cell is in view and then in the snapshot scroll that cell into view by the same amount.
       cssScaling: false,
-      snapshot: env.flags.isEnabled("CLIENT_IM4976_PIVOT_DOWNLOAD"),
+      snapshot: env.flags.isEnabled("CLIENT_IM5878_PIVOT_SNAPSHOT"),
       export: false,
       sharing: false,
       exportData: true,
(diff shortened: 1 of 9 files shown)


src/pivot-table/components/cells/utils/get-view-state.ts
```diff
@@ -0,0 +1,12 @@
+import type { ViewService } from "../../../../types/types";
+
+export const getViewState = (viewService: ViewService) => {
+  if (viewService.viewState) return viewService.viewState;
+  return {
+    rowPartialHeight: 125,
+    visibleTopIndex: 5,
+    visibleRows: 16,
+    page: 0,
+    rowsPerPage: 0,
```
Donya-qlik, line 10: Hard coded for testing the scrolled table snapshot

src/hooks/use-view-service.ts
```diff
@@ -1,8 +1,13 @@
...
+import type { PivotLayout } from "../types/QIX";
 import type { PageInfo, ViewService } from "../types/types";
 
-// eslint-disable-next-line react-hooks/exhaustive-deps
-const useViewService = (pageInfo: PageInfo): ViewService => useMemo(() => createViewService(), [pageInfo.page]);
+const useViewService = (layout: PivotLayout, pageInfo: PageInfo): ViewService =>
+  useMemo(
+    () => createViewService(layout.snapshotData),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [layout.snapshotData, pageInfo.page],
```
Donya-qlik, line 10: I'm not sure pageInfo.page should be in the dependencies. It was there before, so I kept it as it is.

src/ext/index.ts
```diff
@@ -8,7 +8,7 @@ export default function ext(env: Galaxy): Record<string, unknown> {
       // TODO Most of the code is in place to enable snapshot, however WYSIWYG is not always true. Depending on the scroll position, some cell that where visible when the screenshot was taken, is not fully visiable in the screenshot.
       // This should be fixed, possibly be figuring out how much of a cell is in view and then in the snapshot scroll that cell into view by the same amount.
       cssScaling: false,
-      snapshot: false,
+      snapshot: env.flags.isEnabled("CLIENT_IM4976_PIVOT_DOWNLOAD"),
```
Donya-qlik, line 11: Maybe I should create a new flag for the snapshot ..._SNAPSHOT

Donya-qlik (commented):
feat: change the feature flag

src/pivot-table/components/PivotTable.tsx
```diff
@@ -123,6 +127,14 @@ export const StickyPivotTable = ({
   const headerCellRowHightCallback = useCallback(() => headerCellHeight, [headerCellHeight]);
   const contentCellRowHightCallback = useCallback(() => contentCellHeight, [contentCellHeight]);
 
+  useEffect(() => {
+    const element = verticalScrollableContainerRef.current;
+    if (element) {
+      // element.scrollLeft = scrollLeft;
+      element.scrollTop = scrollTopPartially;
```
Donya-qlik, line 134: I definitely need to add a similar property for scrolLeft and other conditions if there are more, but maybe we can focus on vertical scroll for this PR

src/hooks/use-view-service.ts
cbt1: `pageInfo.page` in deps means that the view service is "reset" or re-created when page changes. Which is what we want since all states in viewService are based on the current page.

PR #522, opened 2024-01-22
feat: snapshot scrolling calculations

Donya-qlik:


src/pivot-table/components/cells/utils/snapshot-scroll-calculations.ts
```diff
@@ -0,0 +1,84 @@
...
+  cell: { el: Element; index: number } | undefined,
+  bodyRect: DOMRect | undefined,
+) => {
+  if (!cell) return 0;
+  const cellRect = cell?.el?.getBoundingClientRect();
+  return bodyRect && Math.max(bodyRect.top - cellRect.top, 0);
+};
+
+export const findVisibleRow = (viewService: ViewService, element: HTMLElement) => {
+  const leftColumnBody = element.querySelector(
```
cbt1, line 42: I think you always want to use the "Data grid", since that will always be there even if you don't have any left dimensions

